PR #124, opened 2022-01-08
batch command

alain-rs:
> - [ x] Breaking new feature. If so, please describe details in the description part.

> 
> 
> 
> 
> 
>
> 
> 
> 
> 
> 

> 
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

Pass Interpreter state back and forth batch and master interp

### DIFF
--- a/src/Tcl/TclInterpreter.cpp
+++ b/src/Tcl/TclInterpreter.cpp
@@ -20,9 +20,353 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 #include "TclInterpreter.h"
 
+#include <mutex>
+
 using namespace FOEDAG;
 
 #include <tcl.h>
+
+static std::string TclHistory() {
+  std::string script = R"(
+# history.tcl --
+#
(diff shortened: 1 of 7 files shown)


src/Tcl/TclInterpreter.cpp
```diff
@@ -20,18 +20,364 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
 
+#include <mutex>
+
 using namespace FOEDAG;
 
 #include <tcl.h>
 
+static std::string TclHistory() {
+  std::string script = R"(
+# history.tcl --
```
KochynVolodymyr, line 31: Why not separate file? we will have a big mess with this

alaindargelas: There are a few Tcl files that will need to be sourced with the executable. It is best to embed those files in the executable itself. It will solve a lot of problems with Windows install for instance. Note that we can make those files standalone files. I'll create a separate file for this one.

PR #163, opened 2022-02-05
Console test

alain-rs:

> 
> 
> 
> 
> 
>
> 
> 
> 
> 
> 

> 
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

Console test fails if error during test

### DIFF
--- a/src/Console/Test/console_main.cpp
+++ b/src/Console/Test/console_main.cpp
@@ -38,7 +38,7 @@ void registerExampleCommands(FOEDAG::Session* session) {
     Q_UNUSED(clientData)
     Q_UNUSED(argv)
     Q_UNUSED(argc)
-    FOEDAG::TclConsoleWidget* console = FOEDAG::Init(interp);
+    FOEDAG::TclConsoleWidget* console = FOEDAG::InitConsole(interp);
     QString res = console->getPrompt() + "pwd\n" + QDir::currentPath() + "\n" +
                   console->getPrompt();
     CHECK_EXPECTED("pwd", res)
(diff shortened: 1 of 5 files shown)


KochynVolodymyr:
This changes doesn't solve the problem. If **console_proc** return TCL_ERROR - test still pass.

src/Console/Test/console_main.cpp
```diff
@@ -51,14 +51,12 @@ void registerExampleCommands(FOEDAG::Session* session) {
...
-    QString fileName{"gui_console_proc.tcl"};
-    if (!path.exists(fileName)) return TCL_ERROR;
-    QString fullPath = path.filePath(fileName);
-    FOEDAG::TclConsoleWidget* console = FOEDAG::Init(interp);
-    QString res = console->getPrompt() + "source " + fullPath +
-                  "\nHello world\n" + console->getPrompt();
-    CHECK_EXPECTED("source " + fullPath, res)
+    QFile file(argv[1]);
+    QString expected(argv[2]);
+    if (!file.exists()) return TCL_ERROR;
```
KochynVolodymyr, line 56: if file not exists test will pass successfully.

alain-rs: exit (1) in that case. Please add it.
I checked that if the Tcl code fails or there is a diff, the test fails and the CI fails.

KochynVolodymyr: it still not solve the problem for the whole project. Everyone will expect that **return TCL_ERROR** will cause the CI fail. But it is not. That was the idea of the issue.

PR #165, opened 2022-02-08
suggest tcl testing framework basics

KochynVolodymyr:
Propose tcl testing platform that is similar to gtest.

### DIFF
--- a/src/Console/Test/console_test.cpp
+++ b/src/Console/Test/console_test.cpp
@@ -1,0 +1,61 @@
+/*
+Copyright 2021 The Foedag team
+
+GPL License
+
+Copyright (c) 2021 The Open-Source FPGA Foundation
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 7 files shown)


src/Console/Test/console_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+                 .c_str());
+    return TCL_ERROR;
+  }
+  QString fullPath = file.fileName();
+  FOEDAG::TclConsoleWidget *console = FOEDAG::InitConsole(interp);
+  CHECK_EXPECTED("source " + fullPath, expected)
+  return TCL_OK;
+}
+
+TCL_TEST(console_multiline) {
```
alain-rs, line 55: I'd like to make sure this kind of negative direct testing can still be performed from the outside of the executable (To prevent false positives and false negatives due to bugs in the software itself):
https://github.com/os-fpga/FOEDAG/blob/106acc366711bda2bf14642a01b25fb9cefc3717/Makefile#L134
As long as your new technique is just another method, I'm fine with it.
Please try to use this syntax to capture multi-line scripts in C++:
https://github.com/os-fpga/FOEDAG/blob/106acc366711bda2bf14642a01b25fb9cefc3717/src/Tcl/TclInterpreter.cpp#L74

KochynVolodymyr: So this technique is just another method. In the end it's the same code but I hide some duplicate code under the #define.
I will check false positives and false negatives.

> Please try to use this syntax to capture multi-line scripts in C++:

Why we need this?

alaindargelas:
It's easier to read than having backslash n all over the string. You can build a string with multi lines.

PR #168, opened 2022-02-09
Few updates to testing stuff

KochynVolodymyr:
Improve testing environment.


alain-rs:
Please consider making this generic, not just for the Console, then it should be out of the Console subsystem and in its own subsystem. Not high priority.

PR #210, opened 2022-02-27
Open file in the text editor when user clicked on the file hyperlink …

### LINKED ISSUE #209: Open file in the text editor after hyperlink activated from the console
Open file in the text editor when user clicked on hyperlink in the console.

KochynVolodymyr:
Close #209


KochynVolodymyr:
Here is how it looks like:
image

PR #215, opened 2022-03-01
register tcl command in their respective subsystems

alain-rs:


KochynVolodymyr:
Register tcl commands should be moved outside the modules. This is call dependency injection.

PR #239, opened 2022-03-10
Add tcltest lib (close #235)

### LINKED ISSUE #235: Move out testing library from console and make it separate lib, common for every module
Create TCL testing library. 
Rewrite all tests with tcl testing library.

KochynVolodymyr:
Close #235


src/DesignRuns/Test/design_runs_main.cpp
```diff
@@ -39,32 +40,12 @@ QWidget* DesignRunsBuilder(FOEDAG::CommandLine* cmd,
 }
 
 void registerDesignRunsCommands(QWidget* widget, FOEDAG::Session* session) {
-  auto design_runs_show = [](void* clientData, Tcl_Interp* interp, int argc,
```
alain-rs, line 42: Why are you removing all the registered commands?
They should be usable not only in a test, but end user can write a Tcl script that uses them, or use them interractively in the Tcl console, this comment is valid for all the commands you removed below.

KochynVolodymyr: they are not removed

KochynVolodymyr: they are moved to <...>_test.cpp files

alaindargelas: But the commands are not "test" commands, they are part of the functionality of the GUI.
We write tests that use the commands, but the commands themselves are part of the full functionality of the GUI

KochynVolodymyr: They are test. If you open any TestGui/*.tcl file the only thing we do just run the command and that's it. Nothing else. So why it is not the test?

alain-rs: I think there is one thing that you missed, all the GUI Tcl commands are not only for tests. They can be used by **users**  to create complex macros. All Tcl commands have to be available all the time and accessible from the Tcl interpreter in the GUI.
It looks like your change make those commands only usable in unit tests, which is not correct. Unless I miss something from your changelist.
The minimum requirement is that all commands are always accessible from the interpreter, not just from the unit test.

alain-rs: The Tcl commands will be part of the documentation.

KochynVolodymyr: They are available from the interpreter. As you can see all test pass. No change for end user. Only for developers.

KochynVolodymyr: let's discuss this next time. I see that some of the commands looks like the test but some not. At least for the console I wrote are real tests.

PR #245, opened 2022-03-19
dyn lib for swig loading

alain-rs:
Switch to Tcl 8.6 for SWIG support, also switch to dynamic libraries to enable Tcl loading of shared objects: 
# load <lib.so>


KochynVolodymyr:
Is it possible to split so huge PR? I see there are two main changes - another tcl version and make everything dynamic.

alain-rs:
It's hard enough to get this to work on all platforms.
I will be submitted as a bulk, too many moving parts involved. Also another dependency not visible here yet: SWIG. That is what is motivating this change actually, to be able to load openFPGA as a Tcl dynamic library.

CMakeLists.txt
```diff
@@ -172,20 +172,28 @@ else()
...
 endif()
 
-add_library(tcl_static STATIC IMPORTED)
+link_directories(${CMAKE_CURRENT_BINARY_DIR}/lib/)
+
+get_filename_component(buildDirRelFilePath ${TCL_STATIC_LIB}
+                       REALPATH BASE_DIR ${CMAKE_CURRENT_BINARY_DIR}/lib)
+
+add_library(tcl_static SHARED IMPORTED GLOBAL)
+find_library(tcl_static PATHS ${CMAKE_CURRENT_BINARY_DIR}/lib/)
```
KochynVolodymyr, line 192: this will not work. As I understand find_library is used to find libraries that already installed somewhere in the system, like Qt.

alain-rs: Strangely enough, it works for Ubuntu, Centos and Macos, but does not work for Windows, although I suspect it's the library name that is wrong.

alain-rs:
This changelist 
- introduces dynamic linking for Linux platforms (Ubuntu, CentOS, MacOS) and 
- keep static linking for Windows platforms (Due to not finding the right cmake recipe to make it to work on Windows yet).

The Tcl library dynamic linking is the prime motivation of this PR to enable loading via tcl "load" command swig generated dynamic libraries.

PR #257, opened 2022-03-29
Compiler class

alain-rs:


src/Compiler/Compiler.cpp
```diff
@@ -315,11 +501,19 @@ void Compiler::finish() {
 bool Compiler::RunCompileTask(Action action) {
   switch (action) {
     case Action::Synthesis:
-      return Synthesize();
+      return !Synthesize();
```
KochynVolodymyr, line 504: Why it is inverted here? what is the reason? it looks like bug. It should return true in case success

KochynVolodymyr: Got it!

KochynVolodymyr: It brakes general approaches a little bit. Usually if you have any function that return bool as a success/fail status - true means success. It should be converted in tcl command. I'll create another PR

PR #306, opened 2022-04-25
Redirrect stderr from compiler into the tcl console (close #302)

### LINKED ISSUE #302: stderr shows in the terminal while GUI mode
It should show in the GUI console instead (in red).

alain@alain-xps:~/FOEDAG$ ./dbuild/bin/foedag --compiler openfpga --script tests/Testcases/raygentop/raygentop.tcl 
% Error 1: 
Type: Routing
/home/alain/RapidSilicon/Raptor/OpenFPGA_RS/vpr/src/base/vpr_api.cpp
Line: 801
Message: Fixed channel width must be specified when loading routing (was -1)
Error 1: 
Type: Routing
/home/alain/RapidSilicon/Raptor/OpenFPGA_RS/vpr/src/base/vpr_api.cpp
Line: 801
Message: Fixed channel width must be specified when loading routing (was -1)
done!
QObject::~QObject: Timers cannot be stopped from another thread
Segmentation fault (core dumped)

KochynVolodymyr:
Close #302

### DIFF
--- a/src/Compiler/Compiler.cpp
+++ b/src/Compiler/Compiler.cpp
@@ -27,6 +27,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include <stdlib.h>
 #include <sys/param.h>
 #include <unistd.h>
+
+#include "tclutils/popenRWE.h"
 #endif
 
 #include <QDebug>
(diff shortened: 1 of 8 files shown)


tests/tclutils/popenRWE.cpp
```diff
@@ -0,0 +1,88 @@
...
+  rc = pipe(in);
+  if (rc < 0) goto error_in;
+
+  rc = pipe(out);
+  if (rc < 0) goto error_out;
+
+  rc = pipe(err);
+  if (rc < 0) goto error_err;
+
+  pid = fork();
```
alain-rs, line 35: Absolutely no Fork!
Please remove this component from the source tree. Forks are a terrible idea as they double the process memory size and and not portable.
Please revert back to the Pipe as I coded it.

src/Compiler/Compiler.cpp
```diff
@@ -926,21 +928,33 @@ bool Compiler::ExecuteAndMonitorSystemCommand(const std::string& command) {
...
-  if ((ptr = popen(cmd, "r")) != nullptr) {
-    while (fgets(buf, BUFSIZ, ptr) != nullptr) {
-      if (m_stop == true) {
-        break;
-      }
-      (*m_out) << buf << std::flush;
-    }
-    pclose(ptr);
+  int rwepipe[3];
+  int pid = popenRWE(rwepipe, cmd);
```
alain-rs, line 936: Do not use this component that uses fork

src/Compiler/Compiler.cpp
```diff
@@ -926,21 +928,33 @@ bool Compiler::ExecuteAndMonitorSystemCommand(const std::string& command) {
...
+  if (pid == -1) return false;
+
+  out = fdopen(rwepipe[1], "r");
+  err = fdopen(rwepipe[2], "r");
+  if (!out || !err) return false;
+
+  char *out_ptr, *error_ptr;
+  while (((out_ptr = fgets(buf, BUFSIZ, out)) != nullptr) ||
+         ((error_ptr = fgets(errbuf, BUFSIZ, err)) != nullptr)) {
```
alain-rs, line 945: Simply add this err pulling to the original code

KochynVolodymyr: I need open file descriptor for stderr? how to do it with old approach?

alaindargelas:
Look for a solution that does not involve fork!

tests/tclutils/popenRWE.cpp
KochynVolodymyr: From the **popen()** documentation :
The popen() function opens a process by creating a pipe, forking,
       and invoking the shell.

KochynVolodymyr: this approach is the same as it was before but with error handling

PR #311, opened 2022-04-27
verilog and basic vhdl compilation

alain-rs:


tests/Testcases/trivial/test.tcl
```diff
@@ -0,0 +1,16 @@
...
+add_include_path ./ inc/
+add_design_file test.v bottom.v -SV_2012
+synth
+packing
+place
+route
+sta
+power
+bitstream
+puts "done!"
```
KochynVolodymyr, line 13: sync is missing here

src/Main/main.cpp
```diff
@@ -37,14 +38,21 @@ int main(int argc, char** argv) {
...
+    compiler->SetUseVerific(cmd->UseVerific());
+  } else {
     compiler = new FOEDAG::Compiler();
-  compiler->SetUseVerific(cmd->UseVerific());
+  }
 
   FOEDAG::Foedag* foedag = new FOEDAG::Foedag(
       cmd, mainWindowBuilder, registerAllFoedagCommands, compiler);
-
+  if (opcompiler) {
```
KochynVolodymyr, line 52: Why not pass this values via constructor?

src/Main/main.cpp
```diff
@@ -37,14 +38,21 @@ int main(int argc, char** argv) {
...
+    const std::string& binpath = foedag->Context()->BinaryPath().string();
+    opcompiler->setYosysExecPath(binpath + "/yosys");
```
KochynVolodymyr, line 54: this code required yosis and vrt to be in the same folder as foedag. Why?

src/Main/ToolContext.h
```diff
@@ -39,11 +39,11 @@ class ToolContext {
...
-  const std::string& ExecutableName() { return m_executableName; }
-  const std::filesystem::path& BinaryPath() { return m_binaryPath; }
-  const std::filesystem::path& DataPath() { return m_dataPath; }
+  const std::string& ToolName() const { return m_toolName; }
+  const std::string& CompanyName() const { return m_companyName; }
+  const std::string& ExecutableName() const { return m_executableName; }
+  const std::filesystem::path& BinaryPath() const { return m_binaryPath; }
+  const std::filesystem::path& DataPath() const { return m_dataPath; }
 
   void ToolName(const std::string& name) { m_toolName = name; }
```
KochynVolodymyr, line 48: Why not SetToolName() ? bad code practice

tests/Testcases/trivial/test.tcl
alaindargelas: Added temporarily until you provide a solution that does not require "sync".

src/Main/main.cpp
alaindargelas: See discussion in https://github.com/os-fpga/FOEDAG/pull/314

src/Main/main.cpp
alaindargelas: Because that is the safe way to have the executable you really need. Users might have many versions of Yosys or VPR in the path /usr/bin or /usr/local/bin
This forces the user to create an install that uses the correct Yosys or VPR.
Create symlink of Yosys and VPR to the foedag bin folder.

In the private repo we build all tools and install them in the same directory.

KochynVolodymyr: got it!

src/Main/ToolContext.h
alaindargelas: Google coding style prefers the getters and setters to be the same name:
 void ToolName(const std::string& name) { m_toolName = name; }
 const std::string& ToolName() { return m_toolName; }

KochynVolodymyr: I didn't found this in google codding style. If you know where is it please send some link. I think it isn't a good style.

alain-rs: I'm collaborating with Google employees on another project, that style rule came from there, it is not on their published style sheet but is it mandated or strongly suggested to use internally. Your style vs Google style :-)

KochynVolodymyr: okey

PR #314, opened 2022-04-27
Add compiler factory

KochynVolodymyr:
Code refactoring

### DIFF
--- a/src/Compiler/CMakeLists.txt
+++ b/src/Compiler/CMakeLists.txt
@@ -52,6 +52,7 @@ set (SRC_CPP_LIST
   Task.cpp
   TaskManager.cpp
   CompilerDefines.cpp
+  CompilerFactory.cpp
 )
 
 set (SRC_H_INSTALL_LIST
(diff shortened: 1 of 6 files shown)


alain-rs:
A Compiler factory assumes you know all the compilers.
FOEDAG only contains the basic open-souce compilers.
Each company using FOEDAG will create their own compiler class in their private repos.
The Compiler Factory cannot be used in the general context and will give wrong image that the only 2 compilers are the ones supported by the factory. Also it won't show the right example in the main on how to customize and hook your own compiler.

KochynVolodymyr:
You have --compiler options - it assumes you know all the compilers. What is wrong?

alaindargelas:
A Factory assumes that you have the whole source code of all the possible classes you need to build. 
You don't have the whole source code, example:
You don't have CompilerRapidSilicon.h .cpp class in Foedag, it exists only in private repo.

Each private main.cpp will be unique for each vendor. And they will only have one compiler they support. The --compiler option is just for development purposes. 
An actual main.cpp will not have choices, it will new only one Compiler class.
FOEDAG is a framework it has to show all the flexibility. Nothing (like a Factory) should give the impression that these are the unique possibilites.
Factory is not a good design pattern to use in this circumstance.

PR #317, opened 2022-04-27
VPR channel width option

alain-rs:


tests/Testcases/trivial/test.tcl
```diff
@@ -10,6 +10,7 @@ route
 sta
 power
 bitstream
+sync # Must use for now for proper exit
```
KochynVolodymyr, line 13: I've already removed this command. No Need.

PR #319, opened 2022-04-27
Settings Prototype

skyler-rs:
> Provide a prototype for the Settings system and some example code via the tcl command "DemoSettings"

> An initial interface is provided in main/Settings.cpp. One can set a QJsonValue or QString as the value for a given key. Multiple values can be applied at different hierarchies using updateJson which loops through a QJsonObject and copies each of it's key:value pairs to the system settings.
main/Tasks.cpp provides an example of a system using the settings system to retrieve the tasks info created when reading in settings_test.json

>
> This PR explores a potential settings system built upon Qt's implementation of JSON. Tcl Command "DemoSettings" shows the intended use of loading a prewritten json settings file as well as adding new data to the json structure and reading specific data. Not shown is the concept that user changes should be saved to a separate json file so that their changes can be applied on top of the existing system defaults (this will allow changes to the system defaults w/o a user's settings file overwriting all those updates). The saving user settings concept will be explored when implementing dynamic ui settings generation in the Tasks tab.

Made a draft pull request so there should be no impact


alain-rs:
LGTM, convert it to full PR whenever ready so we can try it.

alain-rs:
@skyler-rs, please update the branch with latest from main before I can merge.

skyler-rs:
@alain-rs updated

src/Main/Settings.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include <QFile>
+#include <QJsonArray>
+#include <QJsonDocument>
+
+#include "stdio.h"
+
+using namespace FOEDAG;
+
+Settings* Settings::m_instance = nullptr;
+Settings::Settings() { fprintf(stderr, "Settings Created\n"); }
```
KochynVolodymyr, line 13: Why not std::cerr ?

src/Main/Settings.h
```diff
@@ -0,0 +1,44 @@
...
+class Settings {
+ private:
+  Settings();
+
+  static Settings* m_instance;
+  QJsonObject m_data;
+
+ public:
+  static Settings* getInstance();
+  static void setValue(QJsonObject& parent, const QString& key,
```
KochynVolodymyr, line 20: what's the point to have all methods static?

src/Main/Tasks.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include <QDebug>
+#include <QJsonArray>
+
+#include "Settings.h"
+
+using namespace FOEDAG;
+
+Tasks* Tasks::m_instance = nullptr;
+Tasks::Tasks() { fprintf(stderr, "Tasks Created\n"); }
```
KochynVolodymyr, line 11: why not std::cerr ?

src/Main/registerTclCommands.cpp
```diff
@@ -213,6 +215,7 @@ void registerBasicBatchCommands(FOEDAG::Session* session) {
   session->TclInterp()->registerCmd("help", help, 0, 0);
 }
 
+#include <QDebug>
```
KochynVolodymyr, line 218: this should be in the include section

CMakeLists.txt
```diff
@@ -459,6 +462,12 @@ add_custom_command(TARGET foedag-bin POST_BUILD
                   COMMAND ${CMAKE_COMMAND} -E copy 
                   ${PROJECT_SOURCE_DIR}/etc/device.xml        
                           ${CMAKE_CURRENT_BINARY_DIR}/share/foedag/etc/)
+                          
+add_custom_command(TARGET foedag-bin POST_BUILD
+                  COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/share/foedag/etc/settings/
+                  COMMAND ${CMAKE_COMMAND} -E copy 
+                  ${PROJECT_SOURCE_DIR}/etc/settings/settings_test.json
```
KochynVolodymyr, line 469: why not add to the resources?

src/Main/Settings.cpp
skyler-rs: Old habit from my previous company which was a bit behind the times and more c styled in their code. This is just debug code anyways.

src/Main/Tasks.cpp
skyler-rs: Just debug code i forgot to remove. Originally thought this PR was going to be more of a temp thing to look at the code instead of an actual check-in. Still getting used to the process ;)

src/Main/registerTclCommands.cpp
skyler-rs: I intend on removing that and the DemoSettings command from the file once the interface is finished up.

CMakeLists.txt
skyler-rs: I don't really know what our directory structure is yet. This is where Alain suggested to put it when i was moving it away from the root folder where I had it when testing/prototyping.

src/Main/Settings.h
KochynVolodymyr: I'm not a fan of singletones, there is always problems when you need write some unit tests. And in general it's bad practice.

CMakeLists.txt
KochynVolodymyr: I mean some resource.qrc file to make settings_test.json as a part of the project

src/Main/Settings.h
skyler-rs: Oh interesting, yeah i'm open to whatever is best. Let's discuss in the the sync up.

skyler-rs: Gave our existing code a look and gave this some more thought. What if I change Settings to a normal class (remove singleton and static elements), and then instantiate during bootstrap and store the object in the Session object similar to how TclInterpreter, CommandLine, CommandStack, etc are doing it? This should address the single instance issues you foresaw while keeping the settings w/ 1 primary copy of the data which I think is important to maintain as more systems begin to read/write to the settings.

CMakeLists.txt
skyler-rs: Ah I understand, this is just a short term file to demo the concept. I envision in the end we will have 1 or more settings definition files that we can read at startup to build our settings. Some of those will be files we deliver with the product and could be stored in a resource file (if the user never needs to see them), some would be files that downstream clients want to add, and finally there are user copies of .json files that contain only their changes to add on top of the settings data. In the end we will probably be accessing these files from multiple locations.

PR #323, opened 2022-04-28
Possible Settings Refactor

skyler-rs:
Per a conversation regarding singleton use in the previous settings PR (https://github.com/os-fpga/FOEDAG/pull/319#discussion_r861167792), I've refactored the settings object to remove singleton and static elements and instead tried to copy the way data is passed around in the Session class.

@alain-rs & @volodymyrkochyn does this look like a better approach to you?

I had to touch a lot more files with this approach which makes me feel like it's a bit more tightly coupled, but wanted to get your opinion.

I also had to re-arrange the order of the last parameters on FOEDAG::Foedag() instantiation as ToolContext is passed w/ a default arg and not provided in a number of our instantiations.

Leaving this as a draft PR for now to make sure this is a desired change.

### DIFF
--- a/src/Main/CMakeLists.txt
+++ b/src/Main/CMakeLists.txt
@@ -125,6 +125,8 @@ install(
     FILES ${PROJECT_SOURCE_DIR}/../Main/CommandLine.h
     FILES ${PROJECT_SOURCE_DIR}/../Main/Foedag.h
     FILES ${PROJECT_SOURCE_DIR}/../Main/ToolContext.h
+    FILES ${PROJECT_SOURCE_DIR}/../Main/Settings.h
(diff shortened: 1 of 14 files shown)


skyler-rs:
Not sure why the build machines are failing on the build... They say Main/Settings.h doesn't exist, but I can see it listed in the file diffs of the PR and this builds fine locally. I'll investigate more tomorrow.

alain-rs:
You did not add that header to the install list of headers, see top CMakeLists.txt for existing example.

Alain Dargelas, Ph.D.
Vice President EDA Software
RapidSilicon

KochynVolodymyr:
That's looks better now.

alain-rs:
Yes,
sudo make install
make test_install

Alain Dargelas, Ph.D.
Vice President EDA Software
RapidSilicon

PR #354, opened 2022-05-06
Improve foedag performance while compile (close #353)

### LINKED ISSUE #353: FOEDAG consumes 100% CPU when Yosys or VPR is running
The sub process uses 100% of a core, but Foedag uses 100% of a second core, it should only use a few percents or 10%.

KochynVolodymyr:
Close #353


alain-rs:
@KochynVolodymyr, That change is not working, it is hanging on several tests/platforms

KochynVolodymyr:
I'll check that

PR #356, opened 2022-05-06
Fixed issue with console when press Start button after previous was c…

KochynVolodymyr:
Steps:
* Run in the console "synth"
* Cancel command
* Press Start button

All output become red

### DIFF
--- a/src/Console/ConsoleInterface.h
+++ b/src/Console/ConsoleInterface.h
@@ -12,9 +12,9 @@ class ConsoleInterface : public QObject {
 
   virtual void run(const QString &command) = 0;
   virtual void abort() = 0;
-  virtual int returnCode() const = 0;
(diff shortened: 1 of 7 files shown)


alain-rs:
@KochynVolodymyr, please update, and rerun that failing step 
 Increment Version / increment-patch-version (pull_request)
@aram-rs, is there a privilege problem with this step?

aram-rs:
> @KochynVolodymyr, please update, and rerun that failing step Increment Version / increment-patch-version (pull_request) @aram-rs, is there a privilege problem with this step?

@alain-rs the step will fail for PRs from Forks as mentioned in [FAQ](https://github.com/EndBug/add-and-commit#working-with-prs), we need to change the checkout step to use [Personal Access Token (PAT)](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) in order to be able to push into Forked repos.

PR #359, opened 2022-05-10
Revert "Improve foedag performance while compile (close #353)"

### LINKED ISSUE #353: FOEDAG consumes 100% CPU when Yosys or VPR is running

alain-rs:
Reverts os-fpga/FOEDAG#357


alain-rs:
This change crashes in  batch mode.
You have to test both batch and GUI mode.

PR #377, opened 2022-05-16
Settings / Widget Factory Refactor using nlohmann Json

skyler-rs:
Updating widget factory to persist widget values for current session.
Reducing settings api for now to see what is actually needed.

Try **EditTaskSettings Placement** (or Routing or Synthesis) in the console to see how widget settings are read from /usr/local/share/foedag/etc/settings/settings_test.json and displayed in a session persistent dialog.

> - [ ] The initial settings implementation was written around QJsonObject. It was later discovered that Qt's implementation of Json isn't sufficient for our needs. Specifically, Qt always returns json data as a copy instead of a reference making it difficult to update existing json and merge settings from various files.
> - [ ] This refactor enables work on the task settings implementation.
> - [ ] Tcl command "EditTaskSettings <taskName>" will demo a dialog loading widget parameters from /usr/local/share/foedag/etc/settings/settings_test.json. The dialog will persist the values for that run (changes aren't saved between runs yet).

> 
> 
> 
> 
> - [ ] Original Settings API entry points have been removed while the actual use of the settings system is determined
> - [ ] Original Qt based json implementation has been replaced with a new json library (https://github.com/nlohmann/json)
> - [ ] The initial format of the settings_test.json example file has changed such that task widgets parameters are stored under a key (some unique object name) in the specific task category instead of in an array which introduced some issues when merging data.
> - [ ] "Visible" was removed from the widget json as it was just an example field that never got implemented. This feature can be added if a real scenario calls for it.
> - [ ] [nlohmann json 3.10.5](https://github.com/nlohmann/json/releases/tag/v3.10.5) has been added to the third party directory


src/Main/Settings.cpp
```diff
@@ -1,144 +1,73 @@
...
-  // Manually update json
-  parent.insert(key, parentObj);
-}
+Settings::Settings() { loadSettings(); }
 
-QJsonValue Settings::get(QJsonObject& parent, const QString& key) {
-  return parent.value(key);
+void Settings::loadSettings() {
+  QStringList jsonFiles{
+      "/usr/local/share/foedag/etc/settings/settings_test.json",
```
KochynVolodymyr, line 33: it is better to hardcore file names outside the class. e.g. when you create object.

src/Main/Settings.h
```diff
@@ -1,39 +1,58 @@
...
-                  QJsonObject& newJson);
-
-  QJsonValue get(QJsonObject& parent, const QString& key);
-  QJsonValue get(const QString& key);
-  QJsonValue getNested(QJsonObject& parent, const QString& keyPath,
-                       const QString& pathSeparator = ".");
-  QJsonValue getNested(const QString& keyPath,
-                       const QString& pathSeparator = ".");
-  QString getJsonStr(const QJsonObject& object);
+  virtual void loadSettings();
```
KochynVolodymyr, line 47: why is virtual? I don't think it is make sense to have some class hierarchy for settings.

src/Main/Settings.cpp
skyler-rs: Good call, I've moved the loadSettings into Foedag.cpp::Foedag() (figured putting it in main.cpp::main() was a bit too low level).

PR #400, opened 2022-05-19
Converted 'enum' to 'enum class'

aram-rs:
The following has been done:

- converted plain `enum` to `enum class` which adds scope to enum values.
- removed generated file.


alaindargelas:
@aram-rs, that is not what we discussed.
These basic synthesize command options need to stay. The other options you plan to implement are compiler specific

synth_option   ......  <<  compiler specific options you are proposing
synthesize delay, area, mixed, none   << Any compiler can implement these options.

This PR is not accaptable.

alaindargelas:
the pnr_options command append to the pack, place, route, options (No option yet, but I'll also add basic generic options for these stages). It should be the same case here.

aram-rs:
Ok, I got the point, then there is no need to change anything in FOEDAG - the `synth_option` compiler specific options will not be added here. However if you want I can keep the enum conversion changes - please let me know.

PR #430, opened 2022-05-31
Adding persistence to task settings dialog system

skyler-rs:
Task tree "Edit Settings" entries now render as buttons and will dynamically load a settings dialog for the parent Task.

Current settings are still test values and aren't integrated with the respective tasks.

If no project is loaded, editing task settings will persist for the session only(currently).

If a project has been loaded, settings will be saved in the project folder under [projectName].settings/ as json files.

When a project is loaded or created, current settings are cleared, system defaults are reloaded and project settings are loaded (see below).

When a project is loaded, all .json files found in the project folder [projectName].settings/ anr loaded and merged into the current settings json.

### DIFF
--- a/src/Compiler/TaskTableView.cpp
+++ b/src/Compiler/TaskTableView.cpp
@@ -47,11 +47,27 @@ TaskTableView::TaskTableView(TaskManager *tManager, QWidget *parent)
 
 void TaskTableView::mousePressEvent(QMouseEvent *event) {
   auto idx = indexAt(event->pos());
-  bool expandAreaClicked = expandArea(idx).contains(event->pos());
-  if (expandAreaClicked) {
-    model()->setData(idx, expandAreaClicked, ExpandAreaRole);
+  auto task = m_taskManager->task(idx.row());
+  bool handled = false;
+
+  if (task && idx.data(Qt::DisplayRole).toString() == "Edit settings") {
+    auto parent = task->parentTask();
+    if (parent != nullptr) {
+      // Handle the event first so the selection moves before the modal dialog
+      QTableView::mousePressEvent(event);
+      handled = true;
+      emit TaskDialogRequested(parent->title());
+    }
+  } else {
+    bool expandAreaClicked = expandArea(idx).contains(event->pos());
+    if (expandAreaClicked) {
+      model()->setData(idx, expandAreaClicked, ExpandAreaRole);
+    }
   }
-  QTableView::mousePressEvent(event);
+  // Handle the event if something else hasn't already
+  if (!handled) {
+    QTableView::mousePressEvent(event);
+  }
 }
 
 void TaskTableView::mouseDoubleClickEvent(QMouseEvent *event) {
@@ -104,7 +120,16 @@ QRect TaskTableView::expandArea(const QModelIndex &index) const {
 void ChildItemDelegate::paint(QPainter *painter,
                               const QStyleOptionViewItem &option,
                               const QModelIndex &index) const {
-  if (index.data(ParentDataRole).toBool()) {
+  if (index.data(Qt::DisplayRole).toString() == "Edit settings") {
+    QStyleOptionButton btn;
+    int padding = 4;
+    btn.rect = option.rect.adjusted(padding, padding, -padding, -padding);
+    btn.text = "Edit Settings";
+    const QWidget *widget = option.widget;
+    QStyle *style = widget ? widget->style() : QApplication::style();
+    style->drawControl(QStyle::CE_PushButton, &btn, painter, widget);
+    return;
+  } else if (index.data(ParentDataRole).toBool()) {
     QStyleOptionViewItem opt = option;
     initStyleOption(&opt, index);
     const QWidget *widget = option.widget;
--- a/src/Compiler/TaskTableView.cpp
+++ b/src/Compiler/TaskTableView.cpp
@@ -48,15 +48,14 @@ TaskTableView::TaskTableView(TaskManager *tManager, QWidget *parent)
 void TaskTableView::mousePressEvent(QMouseEvent *event) {
   auto idx = indexAt(event->pos());
   auto task = m_taskManager->task(idx.row());
-  bool handled = false;
 
-  if (task && idx.data(Qt::DisplayRole).toString() == "Edit settings") {
+  if (task && task->type() == TaskType::Settings) {
     auto parent = task->parentTask();
     if (parent != nullptr) {
       // Handle the event first so the selection moves before the modal dialog
       QTableView::mousePressEvent(event);
-      handled = true;
-      emit TaskDialogRequested(parent->title());
+      emit TaskDialogRequested(task->settingsKey());
+      return;
     }
   } else {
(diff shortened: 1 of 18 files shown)


src/Main/Settings.cpp
```diff
@@ -24,10 +24,24 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include <QFile>
 #include <iostream>
 
+#include "NewProject/ProjectManager/config.h"
+
 using namespace FOEDAG;
 
+#define SETTINGS_DEBUG false
```
KochynVolodymyr, line 31: you can use qDebug() << ""; Looks weird.

src/Main/WidgetFactory.cpp
```diff
@@ -26,11 +26,33 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
 #include <QLabel>
 #include <QMetaEnum>
+#include <iostream>
 
 using namespace FOEDAG;
 
 const QString DlgBtnBoxName{"SettingsDialogButtonBox"};
 
+// Local debug helpers
+#define WIDGET_FACTORY_DEBUG false
```
KochynVolodymyr, line 36: why not qDebug() ?

src/Compiler/TaskTableView.cpp
```diff
@@ -47,11 +47,27 @@ TaskTableView::TaskTableView(TaskManager *tManager, QWidget *parent)
...
-    model()->setData(idx, expandAreaClicked, ExpandAreaRole);
+  auto task = m_taskManager->task(idx.row());
+  bool handled = false;
+
+  if (task && idx.data(Qt::DisplayRole).toString() == "Edit settings") {
+    auto parent = task->parentTask();
+    if (parent != nullptr) {
+      // Handle the event first so the selection moves before the modal dialog
+      QTableView::mousePressEvent(event);
+      handled = true;
```
KochynVolodymyr, line 58: instead of this flag you can just return after emit.

src/Compiler/TaskTableView.cpp
```diff
@@ -47,11 +47,27 @@ TaskTableView::TaskTableView(TaskManager *tManager, QWidget *parent)
 
 void TaskTableView::mousePressEvent(QMouseEvent *event) {
   auto idx = indexAt(event->pos());
-  bool expandAreaClicked = expandArea(idx).contains(event->pos());
-  if (expandAreaClicked) {
-    model()->setData(idx, expandAreaClicked, ExpandAreaRole);
+  auto task = m_taskManager->task(idx.row());
+  bool handled = false;
+
+  if (task && idx.data(Qt::DisplayRole).toString() == "Edit settings") {
```
KochynVolodymyr, line 53: Can you please add one more field to Task - something like Type
emun TaskType {Action, Settings, ...}. 
In this **if** check the task type but not the text.

src/Main/Settings.cpp
skyler-rs: These are rare use debug prints that are only needed when actively developing in this area so i don't want the prints coming out everytime someone is running debug.

src/Main/WidgetFactory.cpp
skyler-rs: These are rare use debug prints that are only needed when actively developing in this area so i don't want the prints coming out everytime someone is running debug.

src/Compiler/TaskTableView.cpp
skyler-rs: Sounds good, I try to avoid mid function returns out of habit, but if that's standard in our project I'll switch over to that.

src/Compiler/TaskTableView.cpp
```diff
@@ -104,7 +120,16 @@ QRect TaskTableView::expandArea(const QModelIndex &index) const {
 void ChildItemDelegate::paint(QPainter *painter,
                               const QStyleOptionViewItem &option,
                               const QModelIndex &index) const {
-  if (index.data(ParentDataRole).toBool()) {
+  if (index.data(Qt::DisplayRole).toString() == "Edit settings") {
```
KochynVolodymyr, line 123: same here with text

src/Compiler/TaskTableView.cpp
KochynVolodymyr: Not sure about standard but we have many returns in the middle of the function

src/Compiler/TaskTableView.cpp
```diff
@@ -47,11 +47,27 @@ TaskTableView::TaskTableView(TaskManager *tManager, QWidget *parent)
...
+  auto task = m_taskManager->task(idx.row());
+  bool handled = false;
+
+  if (task && idx.data(Qt::DisplayRole).toString() == "Edit settings") {
+    auto parent = task->parentTask();
+    if (parent != nullptr) {
+      // Handle the event first so the selection moves before the modal dialog
+      QTableView::mousePressEvent(event);
+      handled = true;
+      emit TaskDialogRequested(parent->title());
```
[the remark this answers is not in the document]
skyler-rs: This was just trying to dynamically work with the existing tableview code. I can look into your suggestions below about a settings type and look into providing a specific category when creating the setting.

src/Compiler/TaskTableView.cpp
skyler-rs: Will do

src/Compiler/TaskTableView.cpp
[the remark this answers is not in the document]
KochynVolodymyr: I'm just curious what happens if we change title? As I understand then we have to change something in the settings and maybe something else. I don't think we have to be depend on the title anywhere. this will cause the problem in the future.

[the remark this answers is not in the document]
skyler-rs: If you changed the parent category title and didn't update the settings.json then it would load an empty json object and display an empty dialog. We can either get it all free dynamically and expect the dev to update their meta data or explicitly define lookup values and require the dev to define those at the beginning. Since i'll be creating a TaskType now I can probably associate that Task w/ a fixed string so that it would survive the dev renaming the categories down the road.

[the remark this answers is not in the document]
KochynVolodymyr: okay

skyler-rs:
Addressed review comments. Please note the following:
For ChildItemDelegate::paint(), there is no access to the task manager unless I add it to the constructor so instead I'm adding  a new TaskTypeRole. I'm currently converting from enum to uint before storing or after reading to avoid adding Q_NAMESPACE
Q_ENUM_NS(TaskType) to the FOEDAG namespace.

Regarding calling setIndexWidget instead of drawing a button in the ChildDelegate paint, I didn't see a good place to call that since the table is mostly constructed from the view. It seems like i'd need to manually scan the tree to find the right indexes to add the button to. Let me know if I'm missing something though.

alain-rs:
@skyler-rs, please resolve conflict with @KochynVolodymyr PR

skyler-rs:
@alain-rs Conflict resolved and the final test run should be finishing shortly

PR #444, opened 2022-06-06
Fixed project open:

KochynVolodymyr:
Fixed project open:
-- add file grouping to project file
-- fixed compiler steps after project opened (problem with paths)

### DIFF
--- a/src/Compiler/Compiler.cpp
+++ b/src/Compiler/Compiler.cpp
@@ -1335,8 +1335,7 @@ int Compiler::ExecuteAndMonitorSystemCommand(const std::string& command) {
   (*m_out) << "Command: " << command << std::endl;
   auto path = std::filesystem::current_path();  // getting path
   (*m_out) << "Path: " << path.string() << std::endl;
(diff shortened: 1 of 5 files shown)


src/Compiler/Compiler.cpp
```diff
@@ -1335,8 +1335,7 @@ int Compiler::ExecuteAndMonitorSystemCommand(const std::string& command) {
   (*m_out) << "Command: " << command << std::endl;
   auto path = std::filesystem::current_path();  // getting path
   (*m_out) << "Path: " << path.string() << std::endl;
-  std::filesystem::current_path(path /
-                                m_projManager->projectName());  // setting path
+  std::filesystem::current_path(m_projManager->projectPath());  // setting path
```
alain-rs, line 1338: Why was this a problem?

KochynVolodymyr: this will not work when you open project.

KochynVolodymyr: Also doesn't work when project created in not default place.

PR #451, opened 2022-06-08
Add mute command key (close #442)

### LINKED ISSUE #442: --mute option
foedag --script <> --batch --mute

to mute the output completely. Log files, artifacts are created and the executable returns a pass/fail status, but the stdout gets muted. Useful for regressions. stderr can still output on screen.

KochynVolodymyr:
Close #442

### DIFF
--- a/src/Main/CommandLine.cpp
+++ b/src/Main/CommandLine.cpp
@@ -77,6 +77,8 @@ void CommandLine::processArgs() {
       m_help = true;
     } else if (token == "--version") {
       m_version = true;
+    } else if (token == "--mute") {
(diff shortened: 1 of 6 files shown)


src/Main/CommandLine.cpp
```diff
@@ -77,6 +77,8 @@ void CommandLine::processArgs() {
       m_help = true;
     } else if (token == "--version") {
       m_version = true;
+    } else if (token == "--mute") {
```
alain-rs, line 80: Please update the help of Compiler and CompilerOpenFPGA

PR #483, opened 2022-06-24
IP generation wip

alain-rs:


alain-rs:
Could not be rebased, had to be squashed

PR #489, opened 2022-06-29
Fixed add_design_files/add_constraint_files, show error if files not …

### LINKED ISSUE #476: add_design_file/add_constraint creates new file if it doesn't exists
**Describe the bug**
add design file from the GUI console creates file if it doesn't exist.

**To Reproduce**
* Open FOEDAG
* run the console:
** create_design asd
** add_design_file asd.v

**Expected behavior**
Command should show error if file doesn't exists. No new created file.

checked commit: 3329563789d190f03163091eb61b984445e35463

KochynVolodymyr:
Close #476

### DIFF
--- a/src/Compiler/Compiler.cpp
+++ b/src/Compiler/Compiler.cpp
@@ -408,10 +408,8 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
         }
         std::filesystem::path the_path = expandedFile;
         if (!the_path.is_absolute()) {
-          expandedFile =
-              std::filesystem::path(compiler->ProjManager()->projectPath() /
-                                    std::filesystem::path("..") / expandedFile)
-                  .string();
+          const auto& path = std::filesystem::current_path();
+          expandedFile = std::filesystem::path(path / expandedFile).string();
         }
         fileList += expandedFile + " ";
       }
@@ -421,7 +419,7 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
                       std::string("\n"));
     if (compiler->m_tclCmdIntegration) {
       std::ostringstream out;
-      bool ok = compiler->m_tclCmdIntegration->TclAddOrCreateDesignFiles(
+      bool ok = compiler->m_tclCmdIntegration->TclAddDesignFiles(
           fileList.c_str(), language, out);
       if (!ok) {
         compiler->ErrorMessage(out.str());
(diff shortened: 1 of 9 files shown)


src/Compiler/Compiler.cpp
```diff
@@ -1499,6 +1494,15 @@ std::string Compiler::ReplaceAll(std::string_view str, std::string_view from,
   return result;
 }
 
+std::pair<std::filesystem::path, int> Compiler::CurrentPath(
```
alain-rs, line 1497: Why do you need to ask Tcl what is the current path?
There is a simple std::filesystem::path API to do so.

alain-rs: All from C++, no need to do a roundtrip to Tcl.

alain-rs: auto path = std::filesystem::current_path();  
If you need to save the current path in the compiler class, you should do that when initializing the class.

KochynVolodymyr: Right! I thought it will not work when do **cd** somewhere in the tcl console but for some reason it works.

src/Main/CompilerNotifier.cpp
```diff
@@ -42,7 +42,7 @@ void CompilerNotifier::notifyStart() {
 
 void CompilerNotifier::notifyFinish() {
   if (m_console) {
-    m_queue.dequeue();
+    if (!m_queue.isEmpty()) m_queue.dequeue();
```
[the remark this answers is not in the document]
KochynVolodymyr: I haven't found the test when I got crash on this. Maybe it was my local changes. 
But any way this **if**  should be here since according to the documentation QQueue::dequeue assumes that container isn't empty.

src/IPGenerate/IPGenerator.cpp
```diff
@@ -212,6 +212,8 @@ bool IPGenerator::Generate() {
     std::string p = expandedFile.string();
     if (!std::filesystem::exists(expandedFile)) {
       std::filesystem::create_directories(expandedFile.parent_path());
+      std::ofstream file{expandedFile};
```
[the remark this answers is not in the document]
KochynVolodymyr: this test Raptor/tests/Testcases/ip_gen_axis_conv/raptor.tcl will fail without this change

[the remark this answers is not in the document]
ravic-rs: This test: https://github.com/os-fpga/FOEDAG/blob/main/Makefile#L165

PR #490, opened 2022-06-30
Utils lib

alain-rs:


src/Utils/StringUtils.h
```diff
@@ -0,0 +1,107 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef FOEDAG_STRINGUTILS_H
+#define FOEDAG_STRINGUTILS_H
+#pragma once
```
KochynVolodymyr, line 24: if you have **#pragma once** you done need header guard:
#ifndef ...
#define ..
#endif

src/Utils/StringUtils.h
```diff
@@ -0,0 +1,107 @@
...
+
+#include <map>
+#include <sstream>
+#include <string>
+#include <string_view>
+#include <vector>
+
+namespace FOEDAG {
+
+class StringUtils final {
```
KochynVolodymyr, line 34: you don't need class here

src/Utils/FileUtils.cpp
```diff
@@ -0,0 +1,145 @@
...
+  std::error_code ec;
+  fs::path fullPath = fs::canonical(path, ec);
+  bool found = (!ec && fileIsRegular(fullPath));
+  if (result != nullptr) {
+    *result = found ? fullPath : path;
+  }
+  return found;
+}
+
+std::string FileUtils::getFileContent(const fs::path& filename) {
```
KochynVolodymyr, line 92: I suggest to change return value as **std::pair<std::string, bool>** in order to understand if there is any error.

src/Utils/FileUtils.cpp
```diff
@@ -0,0 +1,145 @@
...
+      result.append(buffer, in.gcount());
+    }
+  } else {
+    result = "FAILED_TO_LOAD_CONTENT";
+  }
+  return result;
+}
+
+fs::path FileUtils::getPathName(const fs::path& path) {
+  return path.has_parent_path() ? path.parent_path() : "";
```
KochynVolodymyr, line 113: return path.has_parent_path() ? path.parent_path() : fs::path();

src/Utils/FileUtils.cpp
```diff
@@ -0,0 +1,145 @@
...
+  char* dir = nullptr;
+
+  for (dir = strtok(envpath, ":"); dir; dir = strtok(NULL, ":")) {
+    fs::path a_path = std::string(dir) / path;
+    if (FileUtils::fileExists(a_path)) {
+      return a_path;
+    }
+  }
+
+  for (fs::path dir : {"/usr/bin", "/usr/local/bin", "~/.local/bin", "./"}) {
```
KochynVolodymyr, line 135: this will not work for WIN, is it fine?

alain-rs: It will be harmless. We nee to add the valid paths in Windows later.

alain-rs:
I'll ignore all other comments. This code is coming from another project and all of these methods are used as-is.

PR #502, opened 2022-07-12
IP generation wip

alain-rs:


src/Compiler/CompilerOpenFPGA.cpp
```diff
@@ -120,7 +120,18 @@ void CompilerOpenFPGA::Help(std::ostream* out) {
   (*out) << "                                Constraints: set_pin_loc, "
             "set_region_loc, all SDC commands"
          << std::endl;
-  (*out) << "   ipgenerate ?clean?" << std::endl;
+  (*out) << "   add_litex_ip_catalog <directory> : Browses directory for LiteX "
```
volodymyrkochyn, line 123: looks like copy/paste code here

src/Compiler/Compiler.cpp
```diff
@@ -712,6 +723,41 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
   };
   interp->registerCmd("synth_options", synth_options, this, 0);
 
+  auto add_litex_ip_catalog = [](void* clientData, Tcl_Interp* interp, int argc,
```
volodymyrkochyn, line 726: after few month of development this file might have few thousands line of code. it could be moved out of here

PR #505, opened 2022-07-12
Remove valgrind gui_new_file test which is already a regular test

alain-rs:
@KochynVolodymyr, that test should not be a valgrind test in the first place.
This is a functionality test.
Valgrind test are to verify the C++ code is correct (No memory corruption), not to test single every feature of the GUI.
That executable (newfile) must be returning an error code that valgrind does not like, but is memory clean.


ravic-rs:
@alain-rs looking at the checks for this PR, the make test for PR failed. Should we review [the github actions](https://github.com/os-fpga/FOEDAG/runs/7309826335?check_suite_focus=true#step:4:29)?
Also, there's lot of older commits coming in with the merge. It appears to be like what @KochynVolodymyr was referring to in email about rebasing steps.

alaindargelas:
@ravic-rs, the make test failing is the CentOS build, the failure is transient when there are a lot of merges happening or lots of actions at the same time. It typically resolves a PR or action later.

The rebase does not work for me most of the time (Github says you cannot rebase), so merge and squash is the only action allowed.

KochynVolodymyr:
Why it works before? and doesn't work any more?

KochynVolodymyr:
It looks like the real problem just hidden.

PR #519, opened 2022-07-14
Improve error message when file is not exists (close #515)

### LINKED ISSUE #515: Message is not clear enough
Which file does not exist? If there are multiple files on the command line, user don't know which one is missing, we have to give the exact file name that does not exist.
https://github.com/os-fpga/FOEDAG/blob/c22016c97560a79df100d059ffb608c75a16435a/src/ProjNavigator/tcl_command_integration.cpp#L275

KochynVolodymyr:
Provide user more clear and full information which file is not exists for the command add_design_file. Close #515 

Return from the function list of all files that are not exists.

Error before:
image

Error after:
image

### DIFF
--- a/src/NewProject/ProjectManager/project_manager.cpp
+++ b/src/NewProject/ProjectManager/project_manager.cpp
@@ -341,29 +341,32 @@ int ProjectManager::setProjectType(const QString& strType) {
   return ret;
 }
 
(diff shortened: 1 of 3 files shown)


ravic-rs (approved):
Looks good, merge pending with Raptor test.

PR #488, opened 2022-06-27
Adding code for top level settings widget

skyler-rs:
> ### Motivate of the pull request
> -  Adds a top level settings widget that can find and display any settings dialogs that have been saved in settings
> image
> - Add the following to your task/setting dlg json to have it auto added to settings:
`            "_META_": {
                "isSetting": true
            }`
> - Logic has been added to alert the user when navigating away from a settings pane (changing categories in the main settings widget) to prompt them to save changes
> image

This currently doesn't have a menu pick in the UI, but can be invoked with 'EditSettings' in the console or 'EditSettings <task_name>' to auto select the given category.

PR #523, opened 2022-07-14
Fixing a crash introduced by f22bcd0255e3776afaefa36dc64c68a291733c17

skyler-rs:
> - [ ] This fixes a crash introduced by the final commit of https://github.com/os-fpga/FOEDAG/pull/488

> This extends the scope of a try/catch block required by the json library so that the json stays valid during widget creation.
> The commit makes it look larger than it is, the only changes are moving `QString settingName` and `QVBoxLayout* layout` above the try block and then extending the try block to below the call to `widget = FOEDAG::createSettingsWidget(widgetsJson, settingName, tclArgs);`

> This should resolve a crash w/o changing any actual logic, only the scope of the try block is expanding

### DIFF
--- a/src/Main/WidgetFactory.cpp
+++ b/src/Main/WidgetFactory.cpp
@@ -415,50 +415,51 @@ QWidget* FOEDAG::createSettingsPane(const QString& jsonPath,
 
   QWidget* widget = nullptr;
   if (settings) {
+    QString settingName = "";
+    QVBoxLayout* layout = nullptr;
+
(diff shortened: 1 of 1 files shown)


skyler-rs:
The original issue was a crash when clicking edit settings under Tasks > Synthesis and then clicking Ok or Apply
image

This has been manually tested, automated testing of this crash will require additional tcl testing features which are planned for a later date.

ravic-rs (approved):
Looks good.

PR #524, opened 2022-07-15
Issues/518

TarasSydor:
The issue appeared to be table view repaint problem: only part of it was repainted. In a combination with scrolling, it resulted in visual side effects.

### DIFF
--- a/src/Compiler/TaskModel.h
+++ b/src/Compiler/TaskModel.h
@@ -44,6 +44,9 @@ class TaskModel : public QAbstractTableModel {
   TaskManager *taskManager() const;
   void setTaskManager(TaskManager *newTaskManager);
 
+signals:
+  void rowVisibilityChanged(const QModelIndex &topLeft, const QModelIndex &bottomRight);
+
  private:
   bool setData(const QModelIndex &index, const QVariant &value,
                int role = Qt::EditRole) override;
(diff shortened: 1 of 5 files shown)


src/Compiler/TaskModel.h
```diff
@@ -44,6 +44,10 @@ class TaskModel : public QAbstractTableModel {
   TaskManager *taskManager() const;
   void setTaskManager(TaskManager *newTaskManager);
 
+ signals:
+  void rowVisibilityChanged(const QModelIndex &topLeft,
```
KochynVolodymyr, line 48: this is not how it should work. It should go over dataChange(). The question is why it doesn't work with dataChange?

TarasSydor: dataChanged() was emitted only for children of collapsed task, while the parent wasn't in that range and remained not repainted. 
including parent into dataChanged() didn't work because view() immediately hides it. so I added the signal specific to that action, while dataChanged() is now handled only by Qt.

KochynVolodymyr: Include parent should work. Why it doesn't work? what do you mean view() immediately hides it. ?

TarasSydor: On the other hand, we can leave everything as it was and only emit additional layoutChanged() signal, which will make sure everything is repainted. it should solve both issues.

Issue #518, opened 2022-07-14
UI glitch in tasks table view


TarasSydor:
In tasks view, expanding/collapsing 'Packing' and 'Global Placement' tasks happens immediately, but collapse icon gets updated with some delay.

Also, the view may get distorted (rows overlap) as on the screenshot below:
![Screenshot from 2022-07-13 18-49-07](https://user-images.githubusercontent.com/109239890/179017897-d2622fee-0cd9-4c0d-8c6e-70f89ee45416.png)

Steps to reproduce aren't clear: several collapse\expand actions in a row causes visual side effects.

alain-rs:
@TarasSydor, you are added to the list of contributors. You can self-assign now.

PR #525, opened 2022-07-15
trigger table view repaint after collapsing\expanding tasks

TarasSydor:
Both issues are caused by not repainting table view. Triggering it solves both the delay and rows overlap.


ravic-rs (changes requested):
What are the two issues you are referring to here ?
https://github.com/os-fpga/FOEDAG/issues/518 and ?

ravic-rs (approved):
Looks good.

PR #531, opened 2022-07-18
Read constraint file before it added to the project

KochynVolodymyr:
Read constraint file before add it to the project.

We expect that after project build file <filename>_openfpga.sdc should appear in the project folder. It should be non empty.

How to test.
1. Create new project with all files via project wizard and build.
2. Create project and add files from the context menu and build.
3. Open project file and build.


ravic-rs (approved):
Looks good.

Issue #537, opened 2022-07-19
Show the full path of files read by the project 


TarasSydor:
The request originally comes from Avinash.

Currently we only see the file name in highlighted area:
![зображення](https://user-images.githubusercontent.com/109239890/179780945-303d6528-e743-4f7f-8362-9d877b694ecb.png)

Can we show entire path?

The full path is available in files properties panel. Showing complete path in a tab will consume most (if not all) available space and break usability. 
As a solution we (Ravi, Volodymyr, Taras) agreed on showing a tooltip with full path when hovering over the tab. 

As a future improvement we may want to show only part of path (up to 3 branches) or truncate it if width takes more than n pixels.

skyler-rs:
Did a bit of testing w/ this feature earlier in the week and I think tooltips can be added with the following line being added to text_editor_form.cpp::TextEditorForm::OpenFile():
`m_tab_editor->setTabToolTip(index, fileInfo.absoluteFilePath());`

PR #538, opened 2022-07-19
Show full file path in a tab tooltip

TarasSydor:
This commit extends file tabs with a tooltip, containing full file path (#537):

image


ravic-rs:
@TarasSydor was this tested with upstream codebase by loading/creating a project using GUI and tcl? 
Please verify with multiple files.

TarasSydor:
I did GUI testing on updated code based. Tried multiple files in different locations.
Didn't run tcl tests as tooltip only appears when hovering by mouse, which I assume is not catched by tcl.

ravic-rs (approved):
Looks good.

PR #543, opened 2022-07-21
Improve file loading

KochynVolodymyr:
Improve file loading. Current implementation load file on every reopen main window. It doesn't make sense to do it every time. 
I reworked it in a way that load file only when user choose Open Project.

#### What is currently done? (Provide issue link if applicable)
Move open file from reopen main window.

#### What does this pull request change?
We load file only when user open project.

Core

### Impact of the pull request
The P&R View button. This button need to update after file is loaded.

### DIFF
--- a/src/MainWindow/main_window.cpp
+++ b/src/MainWindow/main_window.cpp
@@ -135,11 +135,12 @@ void MainWindow::newProjectDlg() {
 }
 
 void MainWindow::openProject() {
-  QString fileName = "";
+  QString fileName;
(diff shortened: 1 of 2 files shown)


ravic-rs:
@KochynVolodymyr is this ready for review ?

KochynVolodymyr:
@ravic-rs yes if we finish with the release

ravic-rs:
> @ravic-rs yes if we finish with the release

We can still continue merging here.

ravic-rs (approved):
Looks good.

Issue #493, opened 2022-07-02
Move Compiler utility code into src/Utils/


alain-rs:
https://github.com/os-fpga/FOEDAG/blob/a931448f6b2e3df4f1b819a77be3d753632147dd/src/Compiler/Compiler.h#L155
And all the functions below it

PR #545, opened 2022-07-22
Move ProcessUtils files to Utils package. update all includes and cor…

TarasSydor:
Current commit has no functional added value and only moves ProcessUtils.h/cpp files from Compiler to Utils package.

I also updated includes and cmakelists accordingly.

Motivation: #492 issue.


TarasSydor:
The second commit additionaly fixes #493 issue. As it touches the same file, I decided to fix both in a single PR in order to avoid merging conflicts.
trim() functions were already present in StringUtils and implementation was identical, so I just removed the ones from Compiler.
tokenize() was just moved.

ravic-rs (changes requested):
Changes looks good, needs testing with upstream  codebase for with tcl scripts.

KochynVolodymyr (approved):

Issue #466, opened 2022-06-16
Improve Default Project Name Logic in New Project Wizard 


skyler-rs:
> **Describe the bug**
> The New Project wizard defaults to project_1 every time

> **To Reproduce**
> Steps to reproduce the behavior:
> 1. Click File > New Project
> 2. Finished the wizard using the default name project_1
> 3. Restart the app (note this step is only currently required due to https://github.com/os-fpga/FOEDAG/issues/464)
> 4. Click File > New Project
> 5. Observe that the default project name is still project_1

> **Expected behavior**
> Logic should be updated to increment the default project_X number if the project directory already exists, repeat until a unique path is found.

skyler-rs:
Initial gui task for Zaryab once he starts working in the code

PR #549, opened 2022-07-25
Default ProjectName Logic Improved

zaryabkhan-rs:
> - [ ] To address an existing issue, Open --> https://github.com/os-fpga/FOEDAG/issues/466
> - [ ] The New Project wizard defaults to project_1 and project name check action was set on Next Button. Now logic is updated to automatically set the project name which doesn't exist in the directory without pressing Next Button. 

> - When New Project Menu option is clicked, project name automatically sets to new name which doesn't exist in the directory. Each time creating a new project it increments "x" in  project_x as project name. By default it sets to project_1.

>  https://github.com/os-fpga/FOEDAG/issues/466 
>- Project Name updated until a unique path is found without any button clicked.

>  Unique Project Name Automated

> - [ ] It will add flexibility for user to have good user experience.

### DIFF
--- a/src/NewProject/location_form.cpp
+++ b/src/NewProject/location_form.cpp
@@ -22,7 +22,7 @@ locationForm::locationForm(QWidget *parent)
          "Next, you will specify the type of flow you'll be working with. "
          "Finally, you will specify your project sources and choose a default "
          "part."));
-  // int count = 1;
+
   ui->m_labelPname->setText(tr("Project Name:"));
   ui->m_labelPpath->setText(tr("Project Location:"));
   ui->m_checkBox->setText(tr("Create Project Subdirectory"));
--- a/src/NewProject/location_form.cpp
+++ b/src/NewProject/location_form.cpp
@@ -8,6 +8,11 @@ using namespace FOEDAG;
 locationForm::locationForm(QWidget *parent)
     : QWidget(parent), ui(new Ui::locationForm) {
   ui->setupUi(this);
+  int count = 1;
+  QString project_prefix = "project_";
+  while (QDir(project_prefix + QString::number(count)).exists()) {
+    count++;
+  }
   ui->m_labelTitle->setText(tr("Project Location"));
   ui->m_labelCont->setText(
       tr("This wizard will guide you through the creation of a new project."));
@@ -17,22 +22,23 @@ locationForm::locationForm(QWidget *parent)
          "Next, you will specify the type of flow you'll be working with. "
          "Finally, you will specify your project sources and choose a default "
          "part."));
+  // int count = 1;
   ui->m_labelPname->setText(tr("Project Name:"));
   ui->m_labelPpath->setText(tr("Project Location:"));
   ui->m_checkBox->setText(tr("Create Project Subdirectory"));
   ui->m_labelPath0->setText(tr("Project will be created at:"));
   ui->m_btnBrowse->setText(tr("Browse..."));
-  ui->m_lineEditPname->setText(QString("project_") + QString::number(count));
+  ui->m_lineEditPname->setText(project_prefix + QString::number(count));
   ui->m_lineEditPpath->setText(QDir::currentPath());
   ui->m_labelPath1->setText(ui->m_lineEditPpath->text());
   ui->m_checkBox->setCheckState(Qt::CheckState::Checked);
 
-  QString name = ui->m_lineEditPname->text();
-  while (QDir(name).exists()) {
-    ui->m_lineEditPname->setText(QString("project_") +
-                                 QString::number(count++));
-    name = ui->m_lineEditPname->text();
-  };
+  //  QString name = ui->m_lineEditPname->text();
+  //  while (QDir(name).exists()) {
+  //    ui->m_lineEditPname->setText(QString("project_") +
+  //                                 QString::number(count++));
+  //    name = ui->m_lineEditPname->text();
+  //  };
 }
 
 locationForm::~locationForm() { delete ui; }
--- a/src/NewProject/location_form.cpp
+++ b/src/NewProject/location_form.cpp
@@ -3,7 +3,6 @@
 #include <QFileDialog>
 
 #include "ui_location_form.h"
-
 using namespace FOEDAG;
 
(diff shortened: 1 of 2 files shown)


zaryabkhan-rs:
@ravic-rs , Here is the screenshot for project name lineEditBox where we pick the default name and checks if it's exists or not. If project name exists then in "project_x", x increments by 1 and it checks again until a unique name is not found. Once unique name is found then we are good to go.
prnam

ravic-rs:
> @ravic-rs , Here is the screenshot for project name lineEditBox where we pick the default name and checks if it's exists or not. If project name exists then in "project_x", x increments by 1 and it checks again until a unique name is not found. Once unique name is found then we are good to go. prnam

Thanks. Do you think there is an alternative to the prefix "project_"? It sounds too generic.
@skyler-rs what do you think ?

src/NewProject/location_form.h
```diff
@@ -15,6 +15,7 @@ class locationForm : public QWidget {
  public:
   explicit locationForm(QWidget *parent = nullptr);
   ~locationForm();
+  int count = 1;
```
KochynVolodymyr, line 18: you don't need it in header

src/NewProject/location_form.cpp
```diff
@@ -23,10 +22,17 @@ locationForm::locationForm(QWidget *parent)
   ui->m_checkBox->setText(tr("Create Project Subdirectory"));
   ui->m_labelPath0->setText(tr("Project will be created at:"));
   ui->m_btnBrowse->setText(tr("Browse..."));
-  ui->m_lineEditPname->setText("project_1");
+  ui->m_lineEditPname->setText(QString("project_") + QString::number(count));
   ui->m_lineEditPpath->setText(QDir::currentPath());
   ui->m_labelPath1->setText(ui->m_lineEditPpath->text());
   ui->m_checkBox->setCheckState(Qt::CheckState::Checked);
+
+  QString name = ui->m_lineEditPname->text();
```
KochynVolodymyr, line 30: why not:

QString name = QString("project_") + QString::number(count);
while (QDir(name).exists()) {
    name = QString("project_") + QString::number(count++);
  };
ui->m_lineEditPname->setText(name);

skyler-rs:
Hey @zaryabkhan-rs, I had the same comments as Volodymyr. To provide some context for those comments, count is used only one time for determining the project number so a class variable is overkill here; A simple local variable in that function will do fine. For your while loop, you have the correct logic, but each time you assign the value to the UI field which will be done again if the value needs to increment. As such, you can just increment your counter in the loop until you get a valid file name and then you can assign that string to the ui field _once_ afterwards.

For example, I solved the same problem by putting the following at the top of the function:
```
  int count = 1;
  QString project_prefix = "project_";
  while (QDir(project_prefix + QString::number(count)).exists()) {
    count++;  
  }
```
which increments count until project_<count> is an unused project name

and then in the original call to set the field's value i did:
```
ui->m_lineEditPname->setText(project_prefix + QString::number(count));
```

Note that this also somewhat addresses your question about "project_" as a default. I think this is fine since we just need something default that the user can change if they want, but by defining project_prefix and using that variable in our logic, it leaves the code in a state where if we ever did decide to change "project_" we only need to update the line where we set project_prefix.

src/NewProject/location_form.h
zaryabkhan-rs: Yeah! I removed it from here.

src/NewProject/location_form.cpp
zaryabkhan-rs: > Hey @zaryabkhan-rs, I had the same comments as Volodymyr. To provide some context for those comments, count is used only one time for determining the project number so a class variable is overkill here; A simple local variable in that function will do fine. For your while loop, you have the correct logic, but each time you assign the value to the UI field which will be done again if the value needs to increment. As such, you can just increment your counter in the loop until you get a valid file name and then you can assign that string to the ui field _once_ afterwards.
> 
> For example, I solved the same problem by putting the following at the top of the function:
> 
> ```
>   int count = 1;
>   QString project_prefix = "project_";
>   while (QDir(project_prefix + QString::number(count)).exists()) {
>     count++;  
>   }
> ```
> 
> which increments count until project_{count} is an unused project name
> 
> and then in the original call to set the field's value i did:
> 
> ```
> ui->m_lineEditPname->setText(project_prefix + QString::number(count));
> ```
> 
> Note that this also somewhat addresses your question about "project_" as a default. I think this is fine since we just need something default that the user can change if they want, but by defining project_prefix and using that variable in our logic, it leaves the code in a state where if we ever did decide to change "project_" we only need to update the line where we set project_prefix.

Yeah! you are right! I have modified it. Now it looks more consistent. Thankyou for suggestion.

src/NewProject/location_form.h
```diff
@@ -15,11 +15,11 @@ class locationForm : public QWidget {
  public:
   explicit locationForm(QWidget *parent = nullptr);
   ~locationForm();
-
   QString getProjectName();
   QString getProjectPath();
   bool IsCreateDir();
   bool IsProjectNameExit();
+
```
KochynVolodymyr, line 22: can you revert all changes in a header?

src/NewProject/location_form.cpp
```diff
@@ -18,15 +22,23 @@ locationForm::locationForm(QWidget *parent)
...
   ui->m_labelPpath->setText(tr("Project Location:"));
   ui->m_checkBox->setText(tr("Create Project Subdirectory"));
   ui->m_labelPath0->setText(tr("Project will be created at:"));
   ui->m_btnBrowse->setText(tr("Browse..."));
-  ui->m_lineEditPname->setText("project_1");
+  ui->m_lineEditPname->setText(project_prefix + QString::number(count));
   ui->m_lineEditPpath->setText(QDir::currentPath());
   ui->m_labelPath1->setText(ui->m_lineEditPpath->text());
   ui->m_checkBox->setCheckState(Qt::CheckState::Checked);
+
```
KochynVolodymyr, line 35: can you remove commented code?

ravic-rs (approved):
Looks good.

Issue #522, opened 2022-07-14
Status column width mismatch under Task tab


ravic-rs:
> **Describe the bug**
> The status column width within the "Task" tab is not set correctly.

> **To Reproduce**
> Steps to reproduce the behavior:
> 1. Clone repository and checkout commit id: 330f3800b48c (on main branch).
> 2. Build and launch foedag.
> 3. View the Status column under task

> **Expected behavior**
> Status column width should atleast show the entire text "Status" clearly to user. It should behave well when windows are resized.

> **Screenshots**
> ![image](https://user-images.githubusercontent.com/108307503/179117886-d01179c9-ac99-4025-89be-c78e45c4e711.png)

> **Environment (please complete the following information):**
sw03 on server but likely present in all environments.
>  - OS: 
[NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"
CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"]

>  - Compiler:
gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)

> - Version:
 330f3800b48c (on main branch)

ravic-rs:
Please check on the latest commit and verify on upstream GUI as I see it there too.

TarasSydor:
Ravi, I can't reproduce it anymore. Regular startup and with --script command works OK and Status width is correct.
Were there some UI manipulations you did to reproduce it? Tasks view doesn't allow to shrink its width so that columns don't fit.

ravic-rs:
> Ravi, I can't reproduce it anymore. Regular startup and with --script command works OK and Status width is correct. Were there some UI manipulations you did to reproduce it? Tasks view doesn't allow to shrink its width so that columns don't fit.

I can see this both foedag and upstream GUI on the **server**. Please check the build there.

PR #553, opened 2022-07-29
Resize task view columns to their contents instead of explicitly sett…

TarasSydor:
Given change improves tasks view initialization. We ask Qt to initialize column widths to their content in order to avoid truncating some text.

It's supposed to solve #522 issue, but I neither on Ubuntu nor on CentOS could I reproduce it. I theory, with previous approach of setting column widths, it was possible. Current commit should avoid it.

image

### DIFF
--- a/src/Compiler/CompilerDefines.cpp
+++ b/src/Compiler/CompilerDefines.cpp
@@ -43,8 +43,7 @@ QWidget *FOEDAG::prepareCompilerView(Compiler *compiler,
                    FOEDAG::handleTaskDialogRequested);
   view->setModel(model);
 
-  view->setColumnWidth(0, 30);
-  view->setColumnWidth(1, 160);
+  view->resizeColumnsToContents();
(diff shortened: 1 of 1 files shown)


skyler-rs:
I can confirm that this fixes the column:
image
Left side of the image is Ubuntu20.04 via WSL2 without this change, right side is the same sandbox with the change.

One thing to note for future maintenance in this dialog. We should center that column now that it's wider than the icon which is currently left aligned.

ravic-rs:
> I can confirm that this fixes the column: image Left side of the image is Ubuntu20.04 via WSL2 without this change, right side is the same sandbox with the change.
> 
> One thing to note for future maintenance in this dialog. We should center that column now that it's wider than the icon which is currently left aligned.

@TarasSydor this makes sense. Please add open a new GI issue to center the status column.

ravic-rs (approved):
Looks good.

PR #554, opened 2022-07-29
Add progress bar for compilations tasks

KochynVolodymyr:

> 
> 
> 
> 
> 
>
> 
> 
> 
> 
> 

> 
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

### DIFF
--- a/src/Compiler/TaskManager.cpp
+++ b/src/Compiler/TaskManager.cpp
@@ -127,16 +127,18 @@ void TaskManager::startAll() {
   m_runStack.append(m_tasks[TIMING_SIGN_OFF]);
   m_runStack.append(m_tasks[POWER]);
   m_runStack.append(m_tasks[BITSTREAM]);
-  run();
+  m_taskCount = m_runStack.count();
   emit started();
+  run();
 }
 
 void TaskManager::startTask(Task *t) {
   if (!m_runStack.isEmpty()) return;
   if (!t->isValid()) return;
   m_runStack.append(t);
-  run();
+  m_taskCount = m_runStack.count();
   emit started();
+  run();
 }
 
 void TaskManager::startTask(uint id) {
@@ -164,10 +166,19 @@ void TaskManager::runNext() {
     }
   }
 
-  if (m_runStack.isEmpty()) emit done();
+  if (m_runStack.isEmpty()) {
+    m_taskCount = std::nullopt;
+    emit done();
+  }
 }
 
-void TaskManager::run() { m_runStack.first()->trigger(); }
+void TaskManager::run() {
+  if (m_taskCount) {
+    const int max = m_taskCount.value();
+    emit progress(max - m_runStack.count(), max);
+  }
+  m_runStack.first()->trigger();
+}
 
 void TaskManager::reset() {
   for (auto task = m_tasks.begin(); task != m_tasks.end(); task++) {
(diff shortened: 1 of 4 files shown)


KochynVolodymyr:
@skyler-rs it doesn't make sense to have it for one task since it will be from 0 to 100. Sure I can do it but I don't see the point.

KochynVolodymyr:
image

src/Compiler/TaskManager.cpp
```diff
@@ -127,16 +127,18 @@ void TaskManager::startAll() {
   m_runStack.append(m_tasks[TIMING_SIGN_OFF]);
   m_runStack.append(m_tasks[POWER]);
   m_runStack.append(m_tasks[BITSTREAM]);
-  run();
+  m_taskCount = m_runStack.count();
   emit started();
+  run();
```
ravic-rs, line 132: Why was this moved after "emit started()" ?

src/MainWindow/main_window.cpp
```diff
@@ -195,6 +195,19 @@ void MainWindow::createMenus() {
 
   helpMenu = menuBar()->addMenu("&Help");
   helpMenu->addAction(aboutAction);
+
+  m_progressWidget = new QWidget;
```
ravic-rs, line 199: Any reason to not create a separate function for creating this wedget ? e.g., createProgressWidget()

src/Compiler/TaskManager.cpp
KochynVolodymyr: because run() will not exit until task is running

src/MainWindow/main_window.cpp
KochynVolodymyr: Done

src/Compiler/TaskManager.cpp
ravic-rs: I didn't follow why run() was before emit started() before this PR. It seems like we are changing a behavior which I am trying to understand.

KochynVolodymyr: Before this change there was nothing for started() signal. Nothing was connected. No any functionality change

ravic-rs: Thanks, I see the connection now. 
https://github.com/os-fpga/FOEDAG/pull/554/files#diff-9e31125b4e2b6bd449e09a6b478450cb54bf4fb9ddb23baadb06ad5d9a9bb149R402

src/Compiler/TaskManager.h
```diff
@@ -109,6 +114,7 @@ class TaskManager : public QObject {
   QMap<uint, Task *> m_tasks;
   QVector<Task *> m_runStack;
   QMap<Task *, QVector<Task *>> m_rollBack;
+  std::optional<int> m_taskCount{std::nullopt};
```
ravic-rs, line 117: Can we use anything from Qt here for optional value? I see that this is likely the only std we are using in this primarily Qt file.

KochynVolodymyr: I don't see the problem with using anything from the std. This is even better.

KochynVolodymyr: This is the question for discussion but I think we should use std there it is possible.

ravic-rs: I am looking to see if we can separate out Qt vs Std code where we can for testing reasons down the line.

ravic-rs (approved):
Looks good.

PR #560, opened 2022-08-02
IpConfigurator: Add Stub Files

skyler-rs:
> - Milestone 1: check in stub files

> 
> 
> This PR makes the following changes:
> - [ ] Adds Ip Configurator Stub Files
> - [ ] Lays ground work for tcl commands
> - [ ] Adds initial gui smoke tests

> 
> - [ ] GUI
> - [ ] Regression tests

> - [ ] No changes to UI except the instantiation of the IpConfigurator class which will install tcl commands to show/hide the stub ui widget

> ### Testing
> - [ ] As these are only stubs files right now, the testing is limited to tcl smoke tests to ensure that the IpConfigurator class can be built and accessed from the gui
> - [ ] This does its testing on the foedag binary instead of the ipconfigurator binary because the ui needs to do some initialization to fully instantiated the ipconfigurator's subwidgets.


skyler-rs:
Moved the tcl test commands into registerAllFoedagCommands to resolve test failures (initially tried to install the commands into the IpConfigurator widget itself like some other subfolders did, but I believe those folders did their testing on the individual binary created instead of the foedag binary).

ravic-rs (approved):
Looks good. 

@KochynVolodymyr  could you please double-check on the cmake changes? Do we need to make any upstream changes to support these UI changes ?

PR #561, opened 2022-08-03
Show error if target device or architecture files not specified.

KochynVolodymyr:
Show clear message to user that target device or architecture file was not specify.

Check target device and architecture files before synth or place.
Show message in case architecture files were specify but unable to find them.

### DIFF
--- a/src/Compiler/CompilerOpenFPGA.cpp
+++ b/src/Compiler/CompilerOpenFPGA.cpp
@@ -502,9 +502,16 @@ bool CompilerOpenFPGA::RegisterCommands(TclInterpreter* interp,
   return true;
 }
 
+bool CompilerOpenFPGA::VerifyTargetDevice() const {
+  const bool target = Compiler::VerifyTargetDevice();
+  const bool archFile = FileUtils::FileExists(m_architectureFile);
+  return target || archFile;
+}
+
 bool CompilerOpenFPGA::IPGenerate() {
-  PERF_LOG("IPGenerate has started");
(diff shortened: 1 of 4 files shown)


KochynVolodymyr:
@alain-rs  please check if this is expected. We set default architecture file is main.cpp but it will fail if unable to find.

src/Compiler/CompilerOpenFPGA.cpp
```diff
@@ -502,6 +502,14 @@ bool CompilerOpenFPGA::RegisterCommands(TclInterpreter* interp,
   return true;
 }
 
+bool CompilerOpenFPGA::VerifyTargetDevice() const {
+  const bool target = Compiler::VerifyTargetDevice();
+  const bool archFile = FileUtils::FileExists(m_architectureFile);
+  const bool openFpgaArchFile =
+      FileUtils::FileExists(m_OpenFpgaArchitectureFile);
+  return target || (archFile && openFpgaArchFile);
```
alaindargelas, line 510: The test should be return (target || archFile);    for all the stages except bitstream stage.

The openFpgaArchFile is only needed for the bitstream generation stage.

KochynVolodymyr: Ready

alaindargelas (approved):

ravic-rs:
I am looking at how we can test these changes at foedag level with a tcl script, will merge after that.

ravic-rs (approved):
Looks good.

PR #564, opened 2022-08-04
Fixed device selection when new project started

KochynVolodymyr:
This is a fix of a bug that related to new project creation. When user wants create new project, the selection of the device doesn't taking into account.

When user select device and press Ok in the wizard, target_device tcl command is running with corresponding device name. Also, the same happened when file is loading.

 - [ ] Library: <Specify the library name>
 - [x] Engine
 - [ ] Documentation
 - [ ] Regression tests


ravic-rs (approved):
Looks good.

PR #566, opened 2022-08-05
pin assign changes draft

zaryabkhan-rs:

> - Update settings_test.json for placement settings.
> - Add setters/getters for these settings. 
>-  Add pin assignment option in "Add  Constraint file" window and enable the selection flag.
>- Add pin assignment option in Placement Tasks Setting dialog.
>-  Add Tcl API for these settings "pin_loc_assign_method random" and "pin_loc_assign_method DefineOrder.
>- Updated help in FOEDAG/CompilerOpenFPGA.cpp and FOEDAG/Compiler.cpp.

>  This PR improves in the following aspects:  
>- Tcl Interface for placement settings.
>- Pin Assignment Method Selection.
> 
> 

> ### Testing Steps:
>- create_design project_1
>- place Random/DefineOrder- runs the placement test with defined argument.
>- pin_loc_assign_method Random/DefineOrder- change the pin assign selection from console.
>- help- checked the new command added in help for pin assignment.

> 
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

> - [ ] Added new options to GUI for placement settings.

> ### Update Screenshots
image

image

### DIFF
--- a/etc/settings/settings_test.json
+++ b/etc/settings/settings_test.json
@@ -222,14 +222,13 @@
                 "widgetType": "radiobuttons",
                 "options": [
                     "Random",
-                    "In Define Order"
+                    "DefineOrder"
                 ],
                 "optionsLookup":[
                     "Random",
-                    "In Define Order"
-                    
+                    "DefineOrder"
                 ],
-                "layout": "horizontal",
+                "layout": "horizontal", 
                 "arg": "_PlaceOpt_"
             },
             "_META_": {
@@ -259,7 +258,7 @@
                 ],
                 "layout": "horizontal",
                 "default": "b1",
-                "arg": "_placeOpt_"
+                "arg": "_routeOpt_"
             },
             "_META_": {
                 "hidden": false,
--- a/etc/settings/settings_test.json
+++ b/etc/settings/settings_test.json
@@ -217,71 +217,25 @@
             }
         },
         "Placement": {
-            "placement_double_spinbox_ex": {
-                "label": "Double Spinbox Example",
-                "widgetType": "DoubleSpinbox",
-                "minVal": 2.4,
-                "maxVal": 11.4,
-                "stepVal": 1.1,
-                "default": 2.5
-            },
-            "placement_int_spinbox_ex": {
-                "label": "Int Spinbox Example",
-                "widgetType": "Spinbox",
-                "minVal": 2.4,
-                "maxVal": 11.4,
-                "stepVal": 1.1,
-                "default": 2.5
-            },
-            "placement_radiobuttons_vert_ex": {
-                "label": "Radio Button Example (vertical/default)",
+        "placement_radiobuttons_horiz_ex": {
+                "label": "Pin Location Assign Method:",
                 "widgetType": "radiobuttons",
                 "options": [
-                    "b1",
-                    "b2",
-                    "b3"
+                    "Random",
+                    "In Define Order"
                 ],
-                "default": "b2"
-            },
-            "placement_radiobuttons_horiz_ex": {
-                "label": "Radio Button Example (horizontal)",
-                "widgetType": "radiobuttons",
-                "options": [
-                    "b1",
-                    "b2",
-                    "b3"
+                "optionsLookup":[
+                    "Random",
+                    "In Define Order"
+                    
                 ],
                 "layout": "horizontal",
-                "default": "b2"
-            },
-            "placement_checkbox_ex": {
-                "label": "Checkbox Example",
-                "text": "Checkbox Text",
-                "widgetType": "checkbox",
-                "default": "checked"
-            },
-            "placement_dropdown_ex": {
-                "label": "Dropdown Example",
-                "widgetType": "dropdown",
-                "options": [
-                    "option1",
-                    "option2",
-                    "option3"
-                ],
-                "default": "option2"
-            },
-            "placement_input_ex_1": {
-                "label": "Text Input Example",
-                "widgetType": "input",
-                "default": "some text"
-            },
-            "placement_input_ex_2": {
-                "widgetType": "input",
-                "default": "input field w/o a label provided"
+                "arg": "_PlaceOpt_"
             },
             "_META_": {
                 "hidden": false,
-                "isSetting": true
+                "isSetting": true,
+                "tclArgKey": "Tasks_placement"
             }
         },
         "Routing": {
@@ -304,7 +258,8 @@
                     "b3"
                 ],
                 "layout": "horizontal",
-                "default": "b1"
+                "default": "b1",
+                "arg": "_placeOpt_"
             },
             "_META_": {
                 "hidden": false,
--- a/etc/settings/settings_test.json
+++ b/etc/settings/settings_test.json
@@ -217,19 +217,19 @@
             }
         },
         "Placement": {
-        "placement_radiobuttons_horiz_ex": {
+        "pin_selection_radiobtn": {
                 "label": "Pin Location Assign Method:",
                 "widgetType": "radiobuttons",
                 "options": [
                     "Random",
-                    "DefineOrder"
+                    "In Define Order"
                 ],
                 "optionsLookup":[
-                    "Random",
-                    "DefineOrder"
+                    "random",
+                    "in_define_order"
                 ],
-                "layout": "horizontal", 
-                "arg": "_PlaceOpt_"
+                "layout": "horizontal",
+                "arg": "_PinAssign_"
             },
             "_META_": {
                 "hidden": false,
--- a/src/Compiler/Compiler.cpp
+++ b/src/Compiler/Compiler.cpp
@@ -126,13 +126,10 @@ void Compiler::Help(std::ostream* out) {
       << "   synthesize <optimization> ?clean? : Optional optimization (area, "
          "delay, mixed, none)"
       << std::endl;
-  (*out) << "   place <Pin Loc Assign Method> : Pin Loc "
-            "Assign Method "
-            "(DefineOrder(Default), "
-            "Random)"
-         << std::endl;
-  (*out) << "   pin_loc_assign_method <Method>: (DefineOrder(Default)/Random)"
-         << std::endl;
+  (*out) << "   place ?clean" << std::endl;
+  (*out)
+      << "   pin_loc_assign_method <Method>: (in_define_order(Default)/Random)"
+      << std::endl;
   (*out) << "   synth_options <option list>: Synthesis Options" << std::endl;
   (*out) << "   pnr_options <option list>  : PnR Options" << std::endl;
   (*out) << "   packing ?clean?" << std::endl;
@@ -812,10 +809,6 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
         std::string arg = argv[i];
         if (arg == "clean") {
           compiler->PlaceOpt(Compiler::PlacementOpt::Clean);
-        } else if (arg == "Random") {
-          compiler->PlaceOpt(Compiler::PlacementOpt::Random);
-        } else if (arg == "DefineOrder") {
-          compiler->PlaceOpt(Compiler::PlacementOpt::DefineOrder);
         } else {
           compiler->ErrorMessage("Unknown option: " + arg);
         }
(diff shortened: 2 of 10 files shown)


src/Main/Tasks.cpp
```diff
@@ -69,10 +77,21 @@ static std::map<FOEDAG::Compiler::SynthesisOpt, const char*> synthOptMap = {
     {FOEDAG::Compiler::SynthesisOpt::Delay, "delay"},
     {FOEDAG::Compiler::SynthesisOpt::Mixed, "mixed"},
     {FOEDAG::Compiler::SynthesisOpt::Clean, "clean"}};
+
+static std::map<FOEDAG::Compiler::PlacementOpt, const char*> placeOptMap = {
+    {FOEDAG::Compiler::PlacementOpt::None, "none"},
+    {FOEDAG::Compiler::PlacementOpt::Clean, "clean"},
+    {FOEDAG::Compiler::PlacementOpt::Random, "Random"},
```
KochynVolodymyr, line 84: not from upper

src/Main/Tasks.cpp
```diff
@@ -69,10 +77,21 @@ static std::map<FOEDAG::Compiler::SynthesisOpt, const char*> synthOptMap = {
...
+    {FOEDAG::Compiler::PlacementOpt::InDefineOrder, "In Define Order"}};
```
KochynVolodymyr, line 85: I don't think it should have spaces.

src/Main/Tasks.cpp
```diff
@@ -88,7 +107,20 @@ auto synthStrToOpt = [](const QString& str) -> FOEDAG::Compiler::SynthesisOpt {
 
   return val;
 };
+auto placeStrToOpt = [](const QString& str) -> FOEDAG::Compiler::PlacementOpt {
+  auto it = find_if(
+      placeOptMap.begin(), placeOptMap.end(),
+      [str](const std::pair<FOEDAG::Compiler::PlacementOpt, const char*> p) {
+        return p.second == str;
```
KochynVolodymyr, line 114: this will not work. you just compare the pointers but should be string comparison.

src/NewProject/add_constraints_form.ui
```diff
@@ -86,6 +85,59 @@
          </property>
         </widget>
        </item>
+       <item>
```
KochynVolodymyr, line 88: do we need this changes?

src/Main/Tasks.cpp
```diff
@@ -120,6 +164,20 @@ void FOEDAG::TclArgs_setSynthesisOptions(const QString& argsStr) {
...
+  auto [placethArg, moreOpts] = separateArg(PLACE_ARG, argsStr);
+
+  FOEDAG::Compiler* compiler = GlobalSession->GetCompiler();
+  if (compiler) {
+    QStringList tokens = placethArg.split(" ");
+    if (tokens.count() > 1) {
+      int opt = (int)placeStrToOpt(tokens[1]);
+      compiler->PlaceOpt(placeStrToOpt(tokens[1]));
+    }
+    compiler->SynthMoreOpt(moreOpts.toStdString());
```
KochynVolodymyr, line 177: why **SynthMoreOpt** ? looks like typo

zaryabkhan-rs: Yeah, It has been corrected.

src/NewProject/add_constraints_form.ui
zaryabkhan-rs: Yes, Ravi said that Software should also have pin assign option in "Add Constraints" window.

src/Main/Tasks.cpp
zaryabkhan-rs: @KochynVolodymyr at  85 line "In Define Order"?

src/Main/Tasks.cpp
zaryabkhan-rs: I Didn't understand this?

KochynVolodymyr:
Please, put screenshots of the gui in the PR description.

src/Compiler/Compiler.cpp
```diff
@@ -126,6 +126,13 @@ void Compiler::Help(std::ostream* out) {
       << "   synthesize <optimization> ?clean? : Optional optimization (area, "
          "delay, mixed, none)"
       << std::endl;
+  (*out) << "   place <Pin Loc Assign Method> : Pin Loc "
+            "Assign Method "
+            "(DefineOrder(Default), "
```
KochynVolodymyr, line 131: you didn't fix what we have discussed

KochynVolodymyr:
There is different text in the constraint form and setting gui. It supposed to be the same.

etc/settings/settings_test.json
```diff
@@ -217,71 +217,24 @@
...
-                "options": [
-                    "b1",
-                    "b2",
-                    "b3"
-                ],
-                "default": "b2"
-            },
-            "placement_radiobuttons_horiz_ex": {
-                "label": "Radio Button Example (horizontal)",
+        "placement_radiobuttons_horiz_ex": {
```
skyler-rs, line 220: "placement_radiobuttons_horiz_ex" should be renamed to something that describes the setting, make sure to update all references to this as well

etc/settings/settings_test.json
```diff
@@ -217,71 +217,24 @@
...
+                "label": "Pin Location Assign Method:",
                 "widgetType": "radiobuttons",
                 "options": [
-                    "b1",
-                    "b2",
-                    "b3"
+                    "Random",
```
skyler-rs, line 224: Your "options" and "optionsLookup" arrays are the same currently. The "options" field is what is displayed in the ui so it so should be human readable text liked "Define Order", optionsLookup is what will be provided to the back end.

etc/settings/settings_test.json
```diff
@@ -217,71 +217,24 @@
...
-            "placement_input_ex_1": {
-                "label": "Text Input Example",
-                "widgetType": "input",
-                "default": "some text"
-            },
-            "placement_input_ex_2": {
-                "widgetType": "input",
-                "default": "input field w/o a label provided"
+                "layout": "horizontal", 
+                "arg": "_PlaceOpt_"
```
skyler-rs, line 232: This should be the tcl argument that your setting is tied to

src/Main/Tasks.cpp
```diff
@@ -100,7 +133,7 @@ QString FOEDAG::TclArgs_getSynthesisOptions() {
   // SynthMoreOpt so we need to give it a fake arg and pass it
   tclOptions += " -" + QString(SYNTH_ARG) + " " +
                 synthOptToStr(GlobalSession->GetCompiler()->SynthOpt());
-
+  qDebug() << tclOptions;
```
skyler-rs, line 136: remember to remove this for final code

src/NewProject/add_constraints_form.cpp
```diff
@@ -34,3 +37,16 @@ bool addConstraintsForm::IsCopySource() {
   return ui->m_ckkBoxCopy->checkState() == Qt::CheckState::Checked ? true
                                                                    : false;
 }
+
```
skyler-rs, line 40: This code looks problematic. You're assign on top of the parameter being passed in. Either check what is already being reported by the signal by checking define_Order_Checked or if you need to most current value from the widget then just check against ui->select_defineOrder->isChecked(). Same goes for random_checked, the assignment is unneccesary and might actually cause some unintended effects.

src/Compiler/Compiler.cpp
```diff
@@ -805,6 +812,10 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
         std::string arg = argv[i];
         if (arg == "clean") {
           compiler->PlaceOpt(Compiler::PlacementOpt::Clean);
```
skyler-rs, line 814: Is clean a valid option?

KochynVolodymyr: Yes, why?

skyler-rs: The feature being added only has options for Random and DefineOrder so clean seems out of place. I'm wondering if 2 commands are accidentally getting combined here. It seems like there's the Compiler:PlacementOpt which should have Clean and None for the placement task and then I would think we want some separate setting for this pin loc assignment altho I might be misunderstanding how it all combines in the end.

KochynVolodymyr: yes, two commands are combined. You are right.

zaryabkhan-rs: @KochynVolodymyr @skyler-rs  clean option is also tied with place command. forexample, place clean -> clean the placement test. But for the other options those run separately.

skyler-rs: But does you command interact w/ the place/placement command? As far as I understand it, your options are part of the pin_loc_assign_method command so I wouldn't think you'd need to modify the place command (your changes from 999-1044 in compiler.cpp). Maybe i'm missing something?

zaryabkhan-rs: In my understanding, there might be two ways to run place test. If pin assign method i.e., random/in_define_order are supposed to be dependent on place test. then place command should interact with these pin assign commands. As clean is just cleaning the place test which is already done. but if user wants to place either with random pin assignment or in_define_order assignment then it should be like place random/in_define_order that's why I combined both with place as well. So, In short place random/in_define_order will execute the place test with the selected option while
and for pin_loc_assign_method command, it's an Tcl interface for selecting one of the options without running place test, just like we go to Edit Settings... on the gui and select either of the option, which only sets the selected option and then we can just type place command only to run the test. 
Because if we follow the synth command, its options are also interacting with it. Just like when i run synth area, it runs the test with setting the optimization option at "Area".

zaryabkhan-rs: If the requirement is just to select an options and saves the selection either with tcl interface or from the gui action, then I guess only pin_loc_assign_method command is enough.

skyler-rs: Ok i think I'm understanding it now, we can discuss further in the call to make sure it's working as expected, but that logic sounds correct.

zaryabkhan-rs: Yeah Sure!

src/Compiler/Compiler.cpp
```diff
@@ -126,6 +126,10 @@ void Compiler::Help(std::ostream* out) {
       << "   synthesize <optimization> ?clean? : Optional optimization (area, "
          "delay, mixed, none)"
       << std::endl;
+  (*out) << "   place ?clean" << std::endl;
+  (*out)
+      << "   pin_loc_assign_method <Method>: (in_define_order(Default)/Random)"
```
KochynVolodymyr, line 131: should be "random"

src/Main/Tasks.cpp
```diff
@@ -57,8 +59,16 @@ auto separateArg = [](const QString& argName,
       targetArg = argString.mid(argIdx, argString.indexOf("-", argIdx + 1));
       otherArgs = otherArgs.replace(targetArg, "");
     }
-  }
 
+    auto argIdPlace = argString.indexOf("-" + QString(PLACE_ARG));
+    std::cout << "'\nplaceid = " << argIdPlace << endl;
+    //    if (argIdPlace != -1) {
```
KochynVolodymyr, line 65: commented code here

src/Compiler/Compiler.cpp
zaryabkhan-rs: Yeah it just missed. I will fix it.

src/Main/Tasks.cpp
zaryabkhan-rs: sure, I will remove all the comments.

src/Compiler/Compiler.cpp
```diff
@@ -1056,36 +1020,40 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
...
-          compiler->ErrorMessage("Unknown Placement Option: command " + arg);
+          compiler->ErrorMessage("Unknown Placement Option: " + arg);
         }
       };
 
       // If we received a tcl argument
       if (argc > 1) {
         for (int i = 1; i < argc; i++) {
-          setPlaceOption(argv[i]);
+          setPlaceOption(argv[1]);
```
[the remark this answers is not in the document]
zaryabkhan-rs: Yeah, I did it for testing purpose but forget to revert it. I will fix that.

etc/settings/settings_test.json
```diff
@@ -304,7 +257,8 @@
                     "b3"
                 ],
                 "layout": "horizontal",
-                "default": "b1"
+                "default": "b1",
+                "arg": "_routeOpt_"
```
skyler-rs, line 261: This change isn't necessary for your PR

zaryabkhan-rs: alright

src/NewProject/add_constraints_form.cpp
skyler-rs: On second look, on_select_defineOrder_clicked and on_select_random_clicked seem to be unnecessary. The settings radiobuttons widget is already mutually exclusive (when you click 1 radio button, the currently selected option gets de-selected) so you don't need to toggle the two values back and forth. If that's all your logic was intending here then i'd just remove both these and see if your code still works as expected.

zaryabkhan-rs: Yeah You are right. I think should work.

ravic-rs:
@zaryabkhan-rs closing this. Please open a new PR after addressing all the changes discussed. Also, rebase your branch accordingly.

PR #578, opened 2022-08-17
Pin assigment UI skeleton

KochynVolodymyr:
Prepare UI skeleton for Pin Assignment feature

Create necessary widgets and show them on main window. But currently with dummy data.


src/PinAssignment/PinsBaseModel.h
```diff
@@ -19,12 +19,34 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 #pragma once
+#include <QStringList>
+#include <QVector>
 
 namespace FOEDAG {
 
+struct IOPort {
```
skyler-rs, line 27: Any reason why two of these are initialized w/ {} and the others aren't?

skyler-rs (approved):
Looks good. There are some QDockWidget name typos where the item is called Doc instead of Dock. Not sure if it's a big enough issue to change, but might cause some grep frustration if someone searches for Dock in the code etc.
Ex:
portsDocWidget
m_docConsole
packagePinDocWidget
m_pinAssignmentDocs
doc
consoledocwidget

KochynVolodymyr:
> 

You are right. Fixed

PR #601, opened 2022-08-27
Analyze step wip

alain-rs:


alain-rs:
@KochynVolodymyr, the "Run" button of the new task does not send events. The Analyze->Run button does not trigger the new task. Can you check what am I missing?

src/Compiler/CompilerDefines.h
```diff
@@ -51,28 +51,30 @@ Design::Language FromFileType(const QString &type);
...
-static constexpr uint PLACEMENT_TIMING_REPORT{14};
-static constexpr uint ROUTING{15};
-static constexpr uint ROUTING_CLEAN{16};
-static constexpr uint ROUTING_SETTINGS{17};
-static constexpr uint ROUTING_WRITE_NETLIST{18};
-static constexpr uint TIMING_SIGN_OFF{19};
-static constexpr uint POWER{20};
-static constexpr uint BITSTREAM{21};
-static constexpr uint PLACE_AND_ROUTE_VIEW{22};
+static constexpr uint ANALYSIS{1};
```
KochynVolodymyr, line 54: this will brake previous saved files because we save task state into the project file. So, task ID shouldn't be changed. The order in the table is defined in the table model.

KochynVolodymyr:
> @KochynVolodymyr, the "Run" button of the new task does not send events. The Analyze->Run button does not trigger the new task. Can you check what am I missing?

You missed update here: void Compiler::setTaskManager(TaskManager* newTaskManager)

alain-rs:
Thanks, that works, I added the IP at the end. 
Now, the arrows on the tasks ">" collapse one time but they can't be opened again to see the subtasks. Please tell me what needs to be changed
I'm trying to capture in one PR all that is needed to add a new Task.

KochynVolodymyr:
Fixed, please do rebase

PR #602, opened 2022-08-29
Add IO ports table and sync with package pin table

KochynVolodymyr:
Add Ports table for the pin assignment feature. Synchronize selections between ports and package pins

Load ports data from json file and show it in the ports table. Selection in Ports table and Package Pin table are synchronized.

### Screenshots
image

### DIFF
--- a/src/PinAssignment/PortsLoader.cpp
+++ b/src/PinAssignment/PortsLoader.cpp
@@ -20,8 +20,52 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 #include "PortsLoader.h"
 
+#include <QFile>
+#include <iostream>
+
+#include "nlohmann_json/json.hpp"
+using json = nlohmann::ordered_json;
+
 namespace FOEDAG {
 
-PortsLoader::PortsLoader() {}
+PortsLoader::PortsLoader(PortsModel *model, QObject *parent)
+    : QObject(parent), m_model(model) {}
+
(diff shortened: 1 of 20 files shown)


src/PinAssignment/PortsLoader.cpp
```diff
@@ -20,8 +20,52 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+  QFile f{file};
+  if (!f.open(QFile::ReadOnly)) return false;
+
+  QString content = f.readAll();
+  json jsonObject;
+  try {
+    jsonObject = json::parse(content.toStdString());
+  } catch (json::parse_error &e) {
+    // output exception information
+    std::cerr << "Json Error: " << e.what() << '\n'
```
ravic-rs, line 47: Use logger here

ravic-rs (commented):
One comment left for review.

src/PinAssignment/PortsLoader.cpp
KochynVolodymyr: It seems like we don't have infrastructure for logging outside the compiler. At least, I don't have a quick fix

ravic-rs: Sounds good, let's connect offline on this.

ravic-rs (approved):
Looks good. The logging mechanism outside compiler will be discussed separately.

PR #603, opened 2022-08-29
extend add_Design_files command with -work and -l arguments.

TarasSydor:
Given PR extends 'add_Design_files ' TCL command with '-work' and '-L' arguments.
It also makes sure they are passed to Yosys script. So far only Verific parser is supported, as default Yosys one puts all the files from the script into one pile.
UI is also not included here.

By default, empty strings will be used and previous behavior kept.

To make its behavior more clear, here is the example: Input (test.tcl)
...
add_design_file -work lib1 test.v -SV_2012
add_design_file -L lib1 -L lib2 bottom.v -SV_2012
...

Output (test.ys):
...
verific -work lib1  -sv2012 /home/work/Work/FOEDAG-Taras/tests/Testcases/trivial/test.v
verific -L lib1 -L lib2  -sv2012 /home/work/Work/FOEDAG-Taras/tests/Testcases/trivial/bottom.v
...

Tested on FOEDAG application with following cmdline commands:
--verific --compiler openfpga --script /home/work/Work/FOEDAG-Taras/tests/Testcases/trivial/test.tcl

### DIFF
--- a/src/Compiler/Compiler.cpp
+++ b/src/Compiler/Compiler.cpp
@@ -354,6 +354,7 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
     if (argc < 2) {
       compiler->ErrorMessage(
           "Incorrect syntax for add_design_file <file(s)> "
+          "[-work libraryName] [-L libraryName1 -L libraryName2]"
           "<type (-VHDL_1987, -VHDL_1993, -VHDL_2000, -VHDL_2008 (.vhd "
           "default), -V_1995, "
           "-V_2001 (.v default), -SV_2005, -SV_2009, -SV_2012, -SV_2017 (.sv "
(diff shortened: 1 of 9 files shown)


alaindargelas:
", as default Yosys one puts all the files from the script into one pile." Create another PR for this, as I explained, Yosys supports multiple calls to read_verilog but does not support -work or -L (Warning needed)

src/NewProject/ProjectManager/project_fileset.h
```diff
@@ -30,12 +31,17 @@ class ProjectFileSet : public ProjectOption {
   const std::vector<std::pair<QString, QString>> &getMapFiles() const;
   const std::vector<std::pair<int, QStringList>> &Files() const;
 
+  std::vector<std::pair<QStringList, QStringList>> getLibraries() const;
```
KochynVolodymyr, line 34: const reference here

src/Compiler/Compiler.cpp
```diff
@@ -354,6 +354,7 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
     if (argc < 2) {
       compiler->ErrorMessage(
           "Incorrect syntax for add_design_file <file(s)> "
+          "[-work libraryName] [-L libraryName1 -L libraryName2]"
```
KochynVolodymyr, line 357: Need update help as well. For Compiler and CompilerOpenFPGA

TarasSydor:
> ", as default Yosys one puts all the files from the script into one pile." Create another PR for this, as I explained, Yosys supports multiple calls to read_verilog but does not support -work or -L (Warning needed)

Will do. I didn't include it here because -work and -L are optional and won't change the default behavior, while Yosys parser change will affect everyone.

alain-rs:
@TarasSydor please update the PR, @ravic-rs please merge when ready.
I need this change to move forward with https://github.com/os-fpga/FOEDAG/pull/601

PR #604, opened 2022-08-30
Fixed task table, expanding rows issue

KochynVolodymyr:
Issue: task table will not expand row when new rows arrived.

There is wrong conversion between task id and row index.

### Screenshots
image


ravic-rs (approved):
Looks good.

PR #559, opened 2022-08-02
Structure existing unit tests; Rename them according to agreed naming conventions.

TarasSydor:
Given commit doesn't add anything new. It just puts existing unit tests under agreed folder structure and naming conventions.

Changes:
- Command_test.cpp was moved to tests/unittest/Command;
- HelloTcl_test.cpp removed to TclInterpreter_test.cpp and moved to tests/unittest/Tcl;
- CMakeLists test registration updated according to new location.

Updating includes appeared to be needless.

PR #608, opened 2022-08-31
Add QApplication for unittests

KochynVolodymyr:
Allow create unit tests for different kind of widgets.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -264,7 +264,7 @@ jobs:
     - name: Test
       run: |
         make test_install
-        make test/unittest
+        make XVFB="" test/unittest
         make regression
 
   windows-msvc:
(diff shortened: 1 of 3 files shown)


ravic-rs:
@KochynVolodymyr do we have a minimal example of how this going to be used ? I re-call that @TarasSydor was trying something similar in https://github.com/os-fpga/FOEDAG/pull/559 but was removed due to build issues.

KochynVolodymyr:
> @KochynVolodymyr do we have a minimal example of how this going to be used ? I re-call that @TarasSydor was trying something similar in #559 but was removed due to build issues.

This is fix for those build issue. Now we can create unit tests with QWidget, QEvent and so on. E.g. We can create test on this:
https://github.com/os-fpga/FOEDAG/blob/8ee7964b6b9815b93684481a58d5d6c658c512b0/src/PinAssignment/BufferedComboBox.h#L32

ravic-rs (approved):
Looks good.

PR #616, opened 2022-09-02
Fixed crash when welcome page is open

KochynVolodymyr:
There is crash when welcome page started and user try to create new project.

Before welcome page started we need to initialize everything which is done by ReShowWindow method. Not calling it leads to many problems.


ravic-rs (approved):
Looks good.

PR #619, opened 2022-09-02
Fixed broken incremental compilation

KochynVolodymyr:
If project was created from the GUI - incremental build was broken. If user change design file - recompilation haven't happened.

The problem is that we hold wrong path to the files. When add design files we save path to the original and compile them. So the user modification doesn't affect on the design.


ravic-rs:
@KochynVolodymyr we need to wait on tcl test case from Alain before merge.

PR #627, opened 2022-09-07
Add unit tests for the pin assignment module

KochynVolodymyr:
Add unit tests for the pin assignment feature. Also some code improvement and code clean up provided as well.

 - [ ] Library: <Specify the library name>
 - [ ] Engine
 - [ ] Documentation
 - [x] Regression tests

### DIFF
--- a/src/PinAssignment/PackagePinsLoader.cpp
+++ b/src/PinAssignment/PackagePinsLoader.cpp
@@ -29,13 +29,15 @@ PackagePinsLoader::PackagePinsLoader(PackagePinsModel *model, QObject *parent)
 
 PackagePinsLoader::~PackagePinsLoader() {}
 
-bool PackagePinsLoader::load(const QString &fileName) {
-  if (!m_model) return false;
+std::pair<bool, QString> PackagePinsLoader::load(const QString &fileName) {
+  if (!m_model) return std::make_pair(false, "Ports model is missing");
 
   QFile file{fileName};
-  if (!file.open(QFile::ReadOnly)) {
-    return false;
-  }
+  if (!file.exists())
+    return std::make_pair(false,
+                          QString("File %1 doesn't exist").arg(fileName));
+  if (!file.open(QFile::ReadOnly))
+    return std::make_pair(false, QString("Can't open file %1").arg(fileName));
 
   QString content = file.readAll();
 #if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
@@ -64,7 +66,7 @@ bool PackagePinsLoader::load(const QString &fileName) {
   m_model->append(group);  // append last
   m_model->initListModel();
 
-  return true;
+  return std::make_pair(true, QString());
 }
 
 }  // namespace FOEDAG
--- a/src/PinAssignment/PortsLoader.cpp
+++ b/src/PinAssignment/PortsLoader.cpp
@@ -21,7 +21,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "PortsLoader.h"
 
 #include <QFile>
-#include <iostream>
 
 #include "nlohmann_json/json.hpp"
 using json = nlohmann::ordered_json;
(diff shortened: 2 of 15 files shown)


CMakeLists.txt
```diff
@@ -395,44 +393,6 @@ if(MSVC)
   target_compile_options(gtest_main PRIVATE /Zc:__cplusplus)
 endif()
 
-# Concise way to register a new google test
```
ravic-rs, line 398: Why do we no longer need this ? Also, how will it effect the gtests upstream ?

src/PinAssignment/PortsLoader.cpp
```diff
@@ -33,21 +32,23 @@ PortsLoader::PortsLoader(PortsModel *model, QObject *parent)
 
 PortsLoader::~PortsLoader() {}
 
-bool PortsLoader::load(const QString &file) {
-  if (!m_model) return false;
+std::pair<bool, QString> PortsLoader::load(const QString &file) {
+  if (!m_model) return std::make_pair(false, "Ports model is missing");
```
ravic-rs, line 36: Same comment as before.

tests/unittest/PinAssignment/PackagePinsLoader_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+#include "PinAssignment/PackagePinsLoader.h"
+
+#include <filesystem>
+
+#include "PinAssignment/PackagePinsModel.h"
+#include "gtest/gtest.h"
+using namespace FOEDAG;
+
+TEST(PackagePinsLoader, load_general) {
```
ravic-rs, line 30: Please follow the naming convention for gtests.
_When naming your test suites and tests, you should follow the same convention as for [naming functions and classes](https://google.github.io/styleguide/cppguide.html#Function_Names)._
Why ? 
https://google.github.io/googletest/faq.html#why-should-test-suite-names-and-test-names-not-contain-underscore

ravic-rs (changes requested):
Few comments to address.

ravic-rs:
I didn't find the gtests in logs and the log archive file size has reduced significantly.
Previously, they were around ~2MB, this one it's just 160KB.

I suspect it to be a side-effect of removing the gtest registration.
https://github.com/os-fpga/FOEDAG/pull/627#discussion_r966615833

CMakeLists.txt
KochynVolodymyr: Because we reworked gtests and this is useless and do nothing.

src/PinAssignment/PackagePinsLoader.cpp
```diff
@@ -29,13 +29,15 @@ PackagePinsLoader::PackagePinsLoader(PackagePinsModel *model, QObject *parent)
 
 PackagePinsLoader::~PackagePinsLoader() {}
 
-bool PackagePinsLoader::load(const QString &fileName) {
-  if (!m_model) return false;
+std::pair<bool, QString> PackagePinsLoader::load(const QString &fileName) {
+  if (!m_model) return std::make_pair(false, "Ports model is missing");
```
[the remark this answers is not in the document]
KochynVolodymyr: This is something internal that never should happened. what description do you think might be good?

tests/unittest/PinAssignment/PackagePinsLoader_test.cpp
KochynVolodymyr: this is done.

KochynVolodymyr:
> I didn't find the gtests in logs and the log archive file size has reduced significantly. Previously, they were around ~2MB, this one it's just 160KB.
> 
> I suspect it to be a side-effect of removing the gtest registration. [#627 (comment)](https://github.com/os-fpga/FOEDAG/pull/627#discussion_r966615833)

You are right. I don't see reason for this but I remove this change. Now it is good.

src/PinAssignment/PortsLoader.cpp
KochynVolodymyr: This is something internal that never should happened. what description do you think might be good?

tests/unittest/PinAssignment/PackagePinsLoader_test.cpp
ravic-rs: Some tests are still with camelCase as against CamelCase.

KochynVolodymyr: I prefer to see test suit from upper letter like class name and test name from lower case like class method. For me it is more natural. But lets define some documentation on that.

ravic-rs: It's about following the gtest guidelines (defined in their docs already, we don't need a separate one) for naming conventions. Unless we have a good reason to deviate from it we should stick follow them for others who will look at these tests later.

CMakeLists.txt
ravic-rs: This was added back, will be looked in future gtest PR.

src/PinAssignment/PackagePinsLoader.cpp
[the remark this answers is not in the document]
ravic-rs: Ok, it reads as very generic to me but it's a specific failure. Maybe we can try to be specific about the particular error description ?

tests/unittest/PinAssignment/PackagePinsLoader_test.cpp
KochynVolodymyr: Done.

tests/unittest/PinAssignment/PackagePinsLoader_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include "PinAssignment/PackagePinsLoader.h"
+
+#include <filesystem>
+
+#include "PinAssignment/PackagePinsModel.h"
+#include "gtest/gtest.h"
+using namespace FOEDAG;
+
+TEST(PackagePinsLoader, LoadGeneral) {
+  PackagePinsModel model;
```
ravic-rs, line 31: We should start using setup, teardown and fixtures as applicable in this case with LoadAllPins.

ravic-rs (approved):
Looks good.

PR #628, opened 2022-09-07
Extend Set Sources tab with library attributes

TarasSydor:
The motivation of this PR is RG-53.
Previously we extended 'add_design_file' TCL command with RTL support: '-work' and '-L' tags.
This PR also extends new project creation UI with the possibility to enter those libraries:

image

The back end logic, putting the libraries into the Yosys script, is already there.

- Library field accepts String library name and adds it with '-work' tag;
- Import fields accepts comma separated list of library names. Those will be added with -L tag and also duplicated in '-import' command.


ravic-rs (changes requested):
Please add the tcl test like we discussed.

ravic-rs (approved):
Looks good.

PR #631, opened 2022-09-07
Show error message if user selected target device and device size in corresponding vpr.xml doesn't match

KochynVolodymyr:
Add device size validation

When user set device size we validate if it is correct using architecture file. Architecture file must be set before.

 - [x] Library: Compiler
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests


ravic-rs (approved):
Looks good.

PR #634, opened 2022-09-08
generate read_verilog command for each add_design_file call.

TarasSydor:
The motivation of this PR is fixing RG-55 issue.
It's logged on default Yosys parser behavior, which puts the files from all 'add_design_files' commands in a script into a single 'read-verilog'. For example:
add_include_path  inc/
add_design_file bottom.v -SV_2012
add_design_file test.v -SV_2012

Becomes:
read_verilog -sv -I/inc/  bottom.v test.v

With the given fix, each 'add_design_file' call gets corresponding 'read-verilog' command in generated Yosys script.
I also added error message in case -L or -work tags are used with default Yosys parser, as those aren't supported by read_verilog command.
>
The result of tcl script above becomes:
read_verilog -sv -I/inc/  bottom.v
read_verilog -sv -I/inc/  test.v

Based on 'read_verilog' documentation (http://bygone.clairexen.net/yosys/cmd_read_verilog.html), includes have to be populated to each command: _-Idir_ tag.

Tested on Raptor with 'aes_decrypt_fpga' test and on Foedag->trivial (because it adds includes).


ravic-rs (changes requested):
We should be able to execute a tcl test for this behavior and verify the output in both repos. 
Ref:
https://github.com/os-fpga/FOEDAG/blob/main/Makefile#L181
https://wiki.tcl-lang.org/page/grep

KochynVolodymyr (approved):

ravic-rs (approved):
Looks good.

PR #636, opened 2022-09-08
Fix potential empty Available IP tree scenario in Raptor

skyler-rs:
> Discovered an oversight in original Available IP Logic that could result in Raptor displaying an empty IP tree even after a semi successful load of IPs.

> It was assumed that return value of IPCatalogBuilder::buildLiteXCatalog (which loads all IPs at once) would return true if any of the IPs succeeded to load (assuming the return result was OR'd with loop results), but its logic works out such that it returns false if any of the IPs failed to load. My code was overprotectively skipping displaying of any IPs when the return value was false.
>
> The return value is no longer checked and instead the system is always introspected to accurately report what was loaded.

> 
> - [X] Library: IpConfigurator
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

> ### Testing
> - This is a gui only change which can't be tested with our current test harness
> - Manual testing was done in Raptor to ensure that IPs are shown even if one of the IPs fails to load.

### DIFF
--- a/src/IpConfigurator/IpCatalogTree.cpp
+++ b/src/IpConfigurator/IpCatalogTree.cpp
@@ -94,7 +94,7 @@ QStringList IpCatalogTree::getAvailableIPs(
   loadIps(paths);
 
   // Request loaded IPs
-  if (ipsLoaded && tclCmdExists("ip_catalog")) {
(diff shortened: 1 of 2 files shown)


ravic-rs (approved):
Looks good.

PR #638, opened 2022-09-09
make “Show Welcome Page” checkbox persistent between runs

TarasSydor:
The motivation of this change is RG-52 issue, logged on “Show Welcome Page” flag value not stored between runs.
This leads to welcome page appearance after each GUI restart, which might be annoying.

The idea is rather simple: we store an empty file in a working directory. During startup initialization we check whether it's there and don't show the welcome page, if it's not. The file is created when we uncheck "Show Welcome Page" checkbox.
The only way to see the welcome page again is to delete the file from working directory.

### DIFF
--- a/src/MainWindow/main_window.cpp
+++ b/src/MainWindow/main_window.cpp
@@ -56,11 +56,17 @@ extern const char* foedag_version_number;
 extern const char* foedag_git_hash;
 extern const char* foedag_build_type;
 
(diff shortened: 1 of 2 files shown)


skyler-rs (approved):
Looks good.
Note for future considerations if this becomes a more permanent change, we can hide this file (in unix), if we want, by putting a "." in front of the file name

PR #645, opened 2022-09-13
[GEMINIEDA-351] Look for source files in sub directories

TarasSydor:
The motivation of this PR is GEMINIEDA-351 issue.
New project creation wizard has a handy "Add Directories" feature, but it only looks for the files inside that doesn't go into sub directories. This PR removes that limitation. 

Instead of QDir::entryList, which only checks directory content, I used QDirIterator with QDirIterator::Subdirectories flag.
It recursively visits entire directory hierarchy starting from selected one.

Now the user may select high level folder and the code will make sure all the files, satisfying given filter, from all the sub directories are added to the project under creation.

### DIFF
--- a/src/NewProject/source_grid.cpp
+++ b/src/NewProject/source_grid.cpp
@@ -1,5 +1,6 @@
 #include "source_grid.h"
 
+#include <QDirIterator>
 #include <QFileDialog>
 #include <QHeaderView>
 #include <QLabel>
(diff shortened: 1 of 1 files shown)


skyler-rs (approved):
Looks good

src/NewProject/source_grid.cpp
```diff
@@ -148,20 +149,30 @@ void sourceGrid::AddFiles() {
...
+
+  if (folder.isEmpty())  // The dialog was cancelled
+    return;
+  auto it = QDirIterator(folder, {"*.v", "*.sv", "*.vhd"}, QDir::NoFilter,
+                         QDirIterator::Subdirectories);
+  auto files =
+      std::vector<std::pair<QString, QString>>{};  // File names with directory
+                                                   // full paths
+  while (it.hasNext()) {
+    it.next();
```
KochynVolodymyr, line 164: are you sure you don't skip first element here?

TarasSydor: I was also suspecting that, but the first element is kind'a root and is always empty.
Even Qt documentation example works this way:
while (it.hasNext()) {
     qDebug() << it.next();

ravic-rs (approved):

PR #647, opened 2022-09-13
Enable Remove action for top level files; introduce delete shortcut and remove question.

TarasSydor:
The goal of this PR is to solve GEMINIEDA-350 issue.

The changes are:
- Enabled Remove action for top level design file;
- Added a shortcut (Delete) for Remove action;
image

- Ask the user whether he really wants to delete the file.
image

Multi-selection, mentioned in GEMINIEDA-350, was discussed on a meeting and agreed to log it separately, as it's an extra feature that requires some functional analysis.

### DIFF
--- a/src/ProjNavigator/sources_form.cpp
+++ b/src/ProjNavigator/sources_form.cpp
@@ -112,6 +112,7 @@ void SourcesForm::SlotItempressed(QTreeWidgetItem *item, int column) {
                SRC_TREE_SIM_FILE_ITEM == strPropertyRole) {
       if (strName.contains(SRC_TREE_FLG_TOP)) {
         menu->addAction(m_actOpenFile);
+        menu->addAction(m_actRemoveFile);
         menu->addAction(m_actRefresh);
         menu->addSeparator();
         menu->addAction(m_actEditConstrsSets);
@@ -280,9 +281,15 @@ void SourcesForm::SlotRemoveFile() {
   QTreeWidgetItem *item = m_treeSrcHierachy->currentItem();
   if (item == nullptr) return;
 
+  auto strFileName = item->text(0);
+  if (strFileName.contains(SRC_TREE_FLG_TOP))
+    strFileName.remove(SRC_TREE_FLG_TOP);
   auto fileSet = item->data(0, SetFileDataRole);
   if (!fileSet.isNull()) {
-    QString strFileName = item->text(0);
+    auto questionStr = tr("Should %1 file be removed?").arg(strFileName);
+    if (QMessageBox::question(this, tr("Remove file"), questionStr) ==
+        QMessageBox::No)
+      return;
     m_projManager->setCurrentFileSet(fileSet.toString());
     int ret = m_projManager->deleteFile(strFileName);
     if (0 == ret) {
@@ -394,6 +401,8 @@ void SourcesForm::CreateActions() {
           SLOT(SlotRemoveFileSet()));
 
   m_actRemoveFile = new QAction(tr("Remove File"), m_treeSrcHierachy);
+  m_actRemoveFile->setShortcut(Qt::Key_Delete);
+  m_treeSrcHierachy->addAction(m_actRemoveFile);
   connect(m_actRemoveFile, SIGNAL(triggered()), this, SLOT(SlotRemoveFile()));
 
   m_actSetAsTop = new QAction(tr("Set As TopModule"), m_treeSrcHierachy);
@@ -455,6 +464,7 @@ void SourcesForm::CreateFolderHierachyTree() {
           itemf->setIcon(0, QIcon(":/img/file.png"));
           itemf->setData(0, Qt::WhatsThisPropertyRole,
                          SRC_TREE_DESIGN_FILE_ITEM);
+          itemf->setData(0, SetFileDataRole, str);
           parentItem = itemf;
         }
         break;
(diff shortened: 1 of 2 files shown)


skyler-rs (approved):
Looks good.

Note that while doing some testing, i discovered one very random instance where i can crash the program, however this crash occurs with or without Taras' changes so this is a pre-existing issue.

Steps to recreate:
- From FOEDAG dir, run the `./build/bin/foedag --batch --script tests/Testcases/aes_decrypt_fpga/aes_decrypt.tcl` to ensure the AES_DECRYPT folder is available
- Start foedag and click Menu > Open Project
- Load AES_DECRYPT/AES_DECRYPT.ospr
- In the source tree, r-click generic_muxfx.v at the bottom
- Select remove file
For some reason deleting this specific file in this instance causes a crash. I tried adding the file to a different project and it deleted fine so it might be a project issue of some sort.

src/ProjNavigator/sources_form.cpp
```diff
@@ -280,9 +281,15 @@ void SourcesForm::SlotRemoveFile() {
   QTreeWidgetItem *item = m_treeSrcHierachy->currentItem();
   if (item == nullptr) return;
 
+  auto strFileName = item->text(0);
+  if (strFileName.contains(SRC_TREE_FLG_TOP))
+    strFileName.remove(SRC_TREE_FLG_TOP);
   auto fileSet = item->data(0, SetFileDataRole);
   if (!fileSet.isNull()) {
-    QString strFileName = item->text(0);
+    auto questionStr = tr("Should %1 file be removed?").arg(strFileName);
```
KochynVolodymyr, line 289: @ravic-rs please correct text here

TarasSydor: Nice catch Skyler. Actually a dangerous thing is happening in ProjectFileSet::deleteFile:
for (auto it = m_langMap.begin(); it != m_langMap.end(); ++it) {
...
m_langMap.erase(it);
...
}

as that file is the last in the collection, it get invalidated and application crashes. I'll think on another way erasing it.

TarasSydor:
Nice catch Skyler. Actually a dangerous thing is happening in ProjectFileSet::deleteFile:
for (auto it = m_langMap.begin(); it != m_langMap.end(); ++it) {
...
m_langMap.erase(it);
...
}

as that file is the last in the collection, it get invalidated and application crashes. I'll think on another way erasing it.

KochynVolodymyr:
> Nice catch Skyler. Actually a dangerous thing is happening in ProjectFileSet::deleteFile: for (auto it = m_langMap.begin(); it != m_langMap.end(); ++it) { ... m_langMap.erase(it); ... }
> 
> as that file is the last in the collection, it get invalidated and application crashes. I'll think on another way erasing it.

It should be fine. What is the problem with that code?

TarasSydor:
I was too fast with my conclusion. the problem was in the following line - commandLibs container was empty, while it should have the same size as langMap.
operator= lacked one line to also copy command libs.

src/ProjNavigator/sources_form.cpp
```diff
@@ -280,9 +281,15 @@ void SourcesForm::SlotRemoveFile() {
...
+    if (QMessageBox::question(this, tr("Remove file"), questionStr) ==
```
skyler-rs, line 290: Might be able to clarify the remove action by changed the message box title to "Remove file from project?"

src/ProjNavigator/sources_form.cpp
skyler-rs: I think the standard confirmation is "Are you sure" so maybe "Are you sure you want to remove %1 from the project?", I'm open to some other wording for "remove %1 from the project?" if you think there is something better.

ravic-rs (changes requested):
Address comments discussed offline.

src/ProjNavigator/sources_form.cpp
ravic-rs: I'm ok with having it as your proposed.
"Are you sure you want to remote %1 from the project?"
A text below that informs user that we are not deleting the file from disk.
If the user can add this file back using a mechanism then we should provide how to do that as well.
On second thought, I am wondering if we should change the "Remove" to "Delete" instead. Since the above comments with help them understand the difference. The confusion will arise when the shortcut in menu is shown up as "Del"

src/ProjNavigator/sources_form.cpp
ravic-rs: Do we have a general practice to add titles to messageboxes ?

skyler-rs: Could you elaborate on your question? Not sure what is being asked here. QMessageBox always has a title, it's set in line 290 by tr("Remove file").

TarasSydor: I went through message boxes used in our application. They mainly have a title, but not always:
WidgetFactory.cpp:
QMessageBox confirm(parent);
confirm.setWindowTitle("Save Changes?");

main_window.cpp:
auto answer = QMessageBox::question(
        this, "Save constraint file...", ...
        
On the other hand, text_editor_form.cpp has 'titleless' message box (but that might be the only one):
int ret = QMessageBox::question(
        this, tr(""), tr("Save changes in %1?"

ravic-rs: Ok, thanks for sharing that. Let's remove the title in this case and update the sentence as mentioned in comment https://github.com/os-fpga/FOEDAG/pull/647#discussion_r971105075

ravic-rs:
@TarasSydor please update the description to reflect the changes.

TarasSydor:
> @TarasSydor please update the description to reflect the changes.

Done.

ravic-rs:
Typo: "remote"

TarasSydor:
> Typo: "remote" Remove the second line from dialog.: "It can be added...."

question fixed.

PR #652, opened 2022-09-15
Fixed-file not closed in editor after project close action

zaryabkhan-rs:
> - [x] **RG-60 When a file is open in the editor closing the project and creating/opening a new one doesn't close the file/s from previous project or instance**

When a file is open in the editor closing the project and creating/opening a new project doesn't close the file/s from previous project or instance. 
All files opened in the editor as tabs close when Close Project action is invoke from the File Menu. 

Test Steps:
>- Open an example project 
>- Click on the files in source window of main window to open file in an editor.
>- Close Project from the File Menu.
>- Open again an example project, Check the editor as empty. 

> 
> - [x] Library: <Specify the library name>
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests


src/MainWindow/main_window.cpp
```diff
@@ -147,6 +147,13 @@ void MainWindow::SetWindowTitle(const QString& filename, const QString& project,
   }
 }
 
+void MainWindow::CloseFileTabOpened() {
+  auto tabCount = TextEditorForm::Instance()->GetTabWidget()->count();
+  for (auto i = tabCount - 1; i >= 0; i--) {
```
KochynVolodymyr, line 152: why not **for (int i = 0; i < tabCount; i++)** ? this is better for reading

src/MainWindow/main_window.h
```diff
@@ -48,6 +48,7 @@ class MainWindow : public QMainWindow, public TopLevelInterface {
   ProjectInfo Info() const;
   void SetWindowTitle(const QString& filename, const QString& project,
                       QString& projectInfo);
+  void CloseFileTabOpened();
```
KochynVolodymyr, line 51: it is better to rename CloseOpenedTabs()

ravic-rs (changes requested):
Missing details in description of PR.

src/MainWindow/main_window.cpp
zaryabkhan-rs: @KochynVolodymyr  @ravic-rs  The way I implement, When we remove any time it shift the tabs from right to left and index changes depending upon the number of tabs for example, If we have 3 tabs opened in the editor, Index would be 0, 1 2. Now if we removed tab on 0 index, tab on 1 will become index 0 as it will shift to place of 0 index. In our GUI tab shifts from right to left when they close. So it works good in this case.

If we go with suggested logic,  GUI crashes because So When we remove by index in each loop of above loop, it end up removing from index 0,1,2,3 etc, but after we've done  a remove the first time our list has changed.
So if we remove index 0 then everything shifts on index 1 becomes 0 etc, 2 becomes 1 etc, so on our next remove, we are removing at index 1 which is actually now pointing at 2. This continues each time the loop happens. So it crashes whenever tab count increase from more than 1.

src/MainWindow/main_window.h
zaryabkhan-rs: Fixed.

src/MainWindow/main_window.cpp
KochynVolodymyr: can you rework it like this:
`while (count != 0) {...}` ?

src/MainWindow/main_window.cpp
```diff
@@ -147,6 +147,13 @@ void MainWindow::SetWindowTitle(const QString& filename, const QString& project,
   }
 }
 
+void MainWindow::CloseOpenedTabs() {
+  while (TextEditorForm::Instance()->GetTabWidget()->count() != 0) {
```
KochynVolodymyr, line 151: for better readability it is better to do something like this:
`auto tabWidget = TextEditorForm::Instance()->GetTabWidget();`
`while (tabWidget->count() != 0) {...}`

ravic-rs:
@zaryabkhan-rs please merge main into this routinely while PR is under review and also comment on the updates else it will delay the merge.

src/MainWindow/main_window.cpp
zaryabkhan-rs: Fixed.

zaryabkhan-rs:
> @zaryabkhan-rs please merge main into this routinely while PR is under review and also comment on the updates else it will delay the merge.

zaryabkhan-rs:
> @zaryabkhan-rs please merge main into this routinely while PR is under review and also comment on the updates else it will delay the merge.

Fixed!

zaryabkhan-rs:
> Missing details in description of PR.

Issue Title has been added.

KochynVolodymyr (approved):

ravic-rs (approved):
Looks good.

PR #654, opened 2022-09-15
Only disable tasks table viewport and leave scrollbars working

TarasSydor:
This PR aims to solve rg-58 issue: enable scrollbars while compilations is is progress.

I couldn't activate only the scrollbar and leave all the rest as is. It is part of the view which was disabled automatically.
Still, there is an API to retrieve the widget holding only the data part: viewport().
Disabling only that one leaves scrollbars working - exactly what we want.

By doing so other things got enabled either:
- Selection;
- parent tasks collapse/expand;
- column resize.

Those only affect visual representation and, I hope, are acceptible:

image

Note: mouse press events got activated as well, so I had to manually catch them while compilation is in progress. Not the best approach, but it avoids extra complexity and effort to enable scrollbars in another way.

### DIFF
--- a/src/Compiler/TaskTableView.cpp
+++ b/src/Compiler/TaskTableView.cpp
@@ -26,7 +26,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include <QHeaderView>
 #include <QMenu>
 #include <QPushButton>
-#include <QScrollBar>
 
 #include "TaskManager.h"
 
--- a/src/Compiler/TaskTableView.cpp
+++ b/src/Compiler/TaskTableView.cpp
@@ -26,6 +26,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include <QHeaderView>
 #include <QMenu>
 #include <QPushButton>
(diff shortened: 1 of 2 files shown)


src/Compiler/TaskTableView.cpp
```diff
@@ -26,6 +26,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include <QHeaderView>
 #include <QMenu>
 #include <QPushButton>
+#include <QScrollBar>
```
KochynVolodymyr, line 29: I guess this include could be removed

src/Compiler/TaskTableView.h
```diff
@@ -71,5 +71,7 @@ class TaskTableView : public QTableView {
  private:
   TaskManager *m_taskManager{nullptr};
   static constexpr uint TitleCol{1};
+  bool m_viewDisabled{
```
KochynVolodymyr, line 74: Is it possible improve formatting? E.g.
`// comment here`
`bool m_viewDisabled{false};`

ravic-rs (commented):
Looks good.

PR #657, opened 2022-09-16
Continue showing welcome page after user unchecked "Show WelcomePage" checkbox

TarasSydor:
The goal of this PR is to fix RG-64 issue: welcome page immediately disappears after unchecking "Show WelcomePage" checkbox. The user should still be able to execute one of proposed actions.

Not hiding the page immediately opens another opportunity for the user: he may change his mind and check the flag again.
In this case the file has to be removed. Still, all this will have to be reworked after we get QSettings working and decide on configuration panel for this.

image


KochynVolodymyr (approved):

ravic-rs (approved):
Looks good.

PR #658, opened 2022-09-16
[RG-70] Auto load IP Catalog

skyler-rs:
> - [x] To address an existing issue. If so, please provide a link to the issue: RG-70

> Currently the IP_Catalog needs to be manually loaded by calling add_litex_ip_catalog

> add_litex_ip_catalog is now called when ip_catalog and configure_ip are fired and the IP Catalog is empty.

> 
> - [x] Library: IPGenerate, Compiler
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

> - [x] None

> ### Testing
> - tests/TestBatch/test_ip_configure_load.tcl has been added to ensure that an ip can be generated w/o an explicit call to add_litex_ip_catalog
> - Manually integration tested in raptor

### DIFF
--- a/ip_load/axis_converter_V1_0_axis_converter.json
+++ b/ip_load/axis_converter_V1_0_axis_converter.json
@@ -3,7 +3,7 @@
    "core_out_width": 64,
    "core_user_width": 0,
    "core_reverse": 0,
-   "build_dir": "/home/skyler/code/forks/FOEDAG/ipTest/ipTest.IPs/RapidSilicon/IP/axis_converter/V1_1/axis_converter/src",
+   "build_dir": "./ip_load/ip_load.IPs/RapidSilicon/IP/axis_converter/V1_1/axis_converter/src",
    "build_name": "axis_converter.v",
    "build": true,
    "json": "axis_converter_V1_0_axis_converter.json",
--- a/ip_load/axis_converter_V1_0_axis_converter.json
+++ b/ip_load/axis_converter_V1_0_axis_converter.json
@@ -1,0 +1,11 @@
+{
+   "core_in_width": 128,
+   "core_out_width": 64,
+   "core_user_width": 0,
+   "core_reverse": 0,
+   "build_dir": "/home/skyler/code/forks/FOEDAG/ipTest/ipTest.IPs/RapidSilicon/IP/axis_converter/V1_1/axis_converter/src",
+   "build_name": "axis_converter.v",
+   "build": true,
+   "json": "axis_converter_V1_0_axis_converter.json",
+   "json_template": false
+}
--- a/src/Compiler/Compiler.h
+++ b/src/Compiler/Compiler.h
@@ -123,6 +123,8 @@ class Compiler {
   void SetIPGenerator(IPGenerator* generator) { m_IPGenerator = generator; }
   IPGenerator* GetIPGenerator() { return m_IPGenerator; }
   bool BuildLiteXIPCatalog(std::filesystem::path litexPath);
+  bool HasIPInstances();
+  bool HasIPDefinitions();
 
   // VPR, Yosys generic opt
   void ChannelWidth(uint32_t width) { m_channel_width = width; }
@@ -194,7 +196,6 @@ class Compiler {
    */
   virtual bool VerifyTargetDevice() const;
   bool HasTargetDevice();
-  bool HasIPInstances();
 
   bool CreateDesign(const std::string& name);
   static void PrintVersion(std::ostream* out);
--- a/tests/TestBatch/test_ip_configure_load.tcl
+++ b/tests/TestBatch/test_ip_configure_load.tcl
@@ -24,6 +24,7 @@
 
 set platform $::tcl_platform(platform)
 if { $platform == "windows" } {
+    # TODO @skyler-rs Sept2022 Enable in windows once GH-661 is resolved
     puts "SKIPPING ON WINDOWS: This test requires python which FileUtils::ExecuteSystemCommand() currently fails to find on Windows. Disabling windows run of test until python issues on windows are resolved.\n"
     # returning a pass condition because this is an expected failure for now
     exit 0
--- a/tests/TestBatch/test_ip_configure_load.tcl
+++ b/tests/TestBatch/test_ip_configure_load.tcl
@@ -1,0 +1,49 @@
+#Copyright 2022 The Foedag team
+
+#GPL License
+
+#Copyright (c) 2022 The Open-Source FPGA Foundation
+
+#This program is free software: you can redistribute it and/or modify
+#it under the terms of the GNU General Public License as published by
+#the Free Software Foundation, either version 3 of the License, or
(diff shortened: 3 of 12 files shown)


ravic-rs:
Why can't we test this branch in Raptor now ? We should be testing with Raptor before it's merged here.

skyler-rs:
> Why can't we test this branch in Raptor now ? We should be testing with Raptor before it's merged here.
I always put that comment when creating a PR then test in raptor once github creates the PR allowing me to pull the changes into raptor with the `gh pr` command. This one i just forgot to update the comment. This has been integration tested in raptor

src/Compiler/Compiler.h
```diff
@@ -123,6 +123,8 @@ class Compiler {
   void SetIPGenerator(IPGenerator* generator) { m_IPGenerator = generator; }
   IPGenerator* GetIPGenerator() { return m_IPGenerator; }
   bool BuildLiteXIPCatalog(std::filesystem::path litexPath);
+  bool HasIPInstances();
```
KochynVolodymyr, line 126: why not cons?

src/Compiler/Compiler.h
```diff
@@ -123,6 +123,8 @@ class Compiler {
...
+  bool HasIPDefinitions();
```
KochynVolodymyr, line 127: why not const?

src/Compiler/Compiler.h
skyler-rs: Didn't think of it. I tried switching those two to const functions, but they both call GetIPGenerator() which causes a build error once those functions are made const and i don't think we want to make GetIPGenerator const.

src/Compiler/Compiler.h
skyler-rs: Didn't think of it. I tried switching those two to const functions, but they both call GetIPGenerator() which causes a build error once those functions are made const and i don't think we want to make GetIPGenerator const.

ravic-rs:
> > Why can't we test this branch in Raptor now ? We should be testing with Raptor before it's merged here.
> 
> I always put that comment when creating a PR then test in raptor once github creates the PR allowing me to pull the changes into raptor with the `gh pr` command. This one i just forgot to update the comment. This has been integration tested in raptor

Got it, thank you!

ip_load/axis_converter_V1_0_axis_converter.json
```diff
@@ -0,0 +1,11 @@
+{
+   "core_in_width": 128,
+   "core_out_width": 64,
+   "core_user_width": 0,
+   "core_reverse": 0,
+   "build_dir": "./ip_load/ip_load.IPs/RapidSilicon/IP/axis_converter/V1_1/axis_converter/src",
```
ravic-rs, line 6: This is the dummy json generated correct ? The build dir is something we get from ip gen ?

skyler-rs: Yeah this is dummy data until litex dependencies are fixed and we can switch to testing with real IPs. The build dir is set when the python script prints out the generated json, I edited the build_dir path to make it relative instead of hardcoded to my directories.

tests/TestBatch/test_ip_configure_load.tcl
```diff
@@ -0,0 +1,56 @@
...
+
+#You should have received a copy of the GNU General Public License
+#along with this program.  If not, see <http://www.gnu.org/licenses/>.
+######################################################################
+
+# This is a test for GEMINIEDA-352
+# This test ensures that the IP_Catalog is automatically loaded when a user
+# runs a script with configure_ip, but doesn't bother to call add_litex_ip_catalog
+
+set platform $::tcl_platform(platform)
```
ravic-rs, line 25: Can we add a TODO here to fix these with the GH-# ?

tests/Testcases/IPGenerate/IP_Catalog/dummy_generators/RapidSilicon/IP/axis_converter/V1_0/axis_converter_gen.py
```diff
@@ -1,17 +1,21 @@
 #!/usr/bin/env python3
```
ravic-rs, line 1: How are we invoking the this script in linux ? This shebang will only work there.

tests/Testcases/IPGenerate/IP_Catalog/dummy_generators/RapidSilicon/IP/axis_converter/V1_1/axis_converter_gen.py
```diff
@@ -1,17 +1,21 @@
 #!/usr/bin/env python3
```
ravic-rs, line 1: Same question as above.

ravic-rs (commented):
Few comments to address.

tests/TestBatch/test_ip_configure_load.tcl
skyler-rs: done

ip_load/axis_converter_V1_0_axis_converter.json
ravic-rs: Thank you.

tests/Testcases/IPGenerate/IP_Catalog/dummy_generators/RapidSilicon/IP/axis_converter/V1_0/axis_converter_gen.py
skyler-rs: Just a copy paste from [IP_Catalog](https://github.com/RapidSilicon/IP_Catalog/blob/202508e79ef2d0fc37a8513f9d206d1e6322d019/RapidSilicon/IP/axi_ram/v1_0/axi_ram_gen.py#L1) I assume this only allows unix to execute the script w/o specifying the python path, but i think in all our uses we execute the script like "<path to python3> script_gen.py" so I don't think the shebang matters.

tests/Testcases/IPGenerate/IP_Catalog/dummy_generators/RapidSilicon/IP/axis_converter/V1_1/axis_converter_gen.py
skyler-rs: commented above

tests/Testcases/IPGenerate/IP_Catalog/dummy_generators/RapidSilicon/IP/axis_converter/V1_0/axis_converter_gen.py
ravic-rs: Got it, this isn't resolved yet.
The IP Catalog needs to adopt these changes too, it got dropped off my radar last time I saw it.
Please move this issue to pre-jira backlog internally.
https://docs.python.org/3/using/windows.html#python-launcher-for-windows.
I will mark this as resolved now.

PR #659, opened 2022-09-16
Align welcome page controls

TarasSydor:
The goal of this PR is to fix RG-63 issue: labels alignment on the welcome page.

Left margin of description and application name was set to 0, which is similar to actions and recent projects margin.
Show Welcome Page checkbox was moved to the center, as suggested in RG-63.
image


ravic-rs (approved):
Looks good.

PR #667, opened 2022-09-20
[RG-66] Fixed loading project file

KochynVolodymyr:
Project file loading has a critical issue related to incorrect loading.

Set target device causes rewrite project file before it is actually loading. Move this to correct place.

 - [x] Library: foedagcore
 - [x] Engine
 - [ ] Documentation
 - [ ] Regression tests

### DIFF
--- a/src/Main/ProjectFile/ProjectFileLoader.cpp
+++ b/src/Main/ProjectFile/ProjectFileLoader.cpp
@@ -49,6 +49,12 @@ void ProjectFileLoader::registerComponent(ProjectFileComponent *comp,
 }
 
 void ProjectFileLoader::Load(const QString &filename) {
+  m_loadDone = false;
+  LoadInternal(filename);
+  m_loadDone = true;
+}
+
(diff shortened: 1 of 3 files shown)


src/Main/ProjectFile/ProjectFileLoader.cpp
```diff
@@ -82,6 +82,15 @@ void ProjectFileLoader::Load(const QString &filename) {
     }
   }
 
+  // set device
```
ravic-rs, line 85: How's this move solving the behavior reported on the ticket ? Like why is it not set to None now.

ravic-rs (commented):
One comment to address.

src/Main/ProjectFile/ProjectFileLoader.cpp
KochynVolodymyr: This code causes save file during loading. Just before task state loading. It is very complicated and better to have a call for explanation. But I put additional protection against this.

PR #668, opened 2022-09-20
[RG-83] Recognize *.VHD and *.SV files in case insensitive manner

TarasSydor:
This PR fixes first part of RG-83 issue: make sure *.VHD and *.SV files, passed to add_design_files TCL function, are recognized in case insensitive manner.

There are two ways of detecting which language to use: 
- explicitly indicate it as a parameter of add_design_files():
      add_design_file aes_decrypt128.sv aes_decrypt256.sv  -SV_2012

- Or the application automatically deduces the language. Currently only two formats are handled explicitly:
    VHDL_2008 (*.vhd)
    SYSTEMVERILOG_2017 (*.sv)

Automatic language detection was case sensitive: if the file had *.VHD type, it wasn't recognized as VHDL_2008.
This PR makes sure type is recognized in case insensitive manner. We don't override any files or introduce new types.

> 
> - [+ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

Now following input:  add_design_file aes_decrypt128.SV aes_decrypt256.SV
Results in:                 verific -sv aes_decrypt128.SV aes_decrypt256.SV

### DIFF
--- a/src/Compiler/Compiler.cpp
+++ b/src/Compiler/Compiler.cpp
@@ -376,12 +376,12 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
     }
     std::string actualType = "VERILOG_2001";
     Design::Language language = Design::Language::VERILOG_2001;
-    const std::string file = argv[1];
-    if (strstr(file.c_str(), ".vhd")) {
+    auto file = QString(argv[1]).toLower();
+    if (file.contains(".vhd")) {
       language = Design::Language::VHDL_2008;
       actualType = "VHDL_2008";
     }
-    if (strstr(file.c_str(), ".sv")) {
+    if (file.contains(".sv")) {
       language = Design::Language::SYSTEMVERILOG_2017;
       actualType = "SV_2017";
     }
--- a/src/Compiler/Compiler.cpp
+++ b/src/Compiler/Compiler.cpp
@@ -56,6 +56,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "TaskManager.h"
 #include "Utils/FileUtils.h"
 #include "Utils/ProcessUtils.h"
+#include "Utils/StringUtils.h"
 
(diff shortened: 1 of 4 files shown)


src/Compiler/Compiler.cpp
```diff
@@ -376,12 +376,12 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
     }
     std::string actualType = "VERILOG_2001";
     Design::Language language = Design::Language::VERILOG_2001;
-    const std::string file = argv[1];
-    if (strstr(file.c_str(), ".vhd")) {
+    auto file = QString(argv[1]).toLower();
```
KochynVolodymyr, line 379: Compiler must use std::string only. This is an agreement. I suggest to write some function **StringUtils::toLower**

src/Compiler/Compiler.cpp
```diff
@@ -376,12 +376,12 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
...
+    if (file.contains(".vhd")) {
```
KochynVolodymyr, line 380: need to check if is the only place that have this problem.

KochynVolodymyr: https://github.com/os-fpga/FOEDAG/blob/5d94adc66230aeade97d55630bbc61aa5a881963/src/Compiler/Compiler.cpp#L492

src/Compiler/Compiler.cpp
TarasSydor: fixed. also added unit tests for new utility functions

src/Compiler/Compiler.cpp
TarasSydor: covered in second commit. also for .eblif

KochynVolodymyr (approved):

ravic-rs (approved):
Looks good.

PR #680, opened 2022-09-23
[RG-19 - RG-23] Update New Project Dialog UI

TarasSydor:
This PR updates the UI of New Project Dialog. It fixes RG-19, RG-20, RG-21, RG-22, RG-23.

The main change is creating tabs instead of separate screens. To achieve it, QStackedWidget was replaced by QTabWidget.
Standard control has some limitations in terms of tab styling. In order to get desired behavior (text alignment and label direction) a custom style was implemented (taken from stack overflow :) ).

Other changes are rather cosmetic. Here is the complete list;
- Separate screens replaced by tabs. We agreed that tabs won't be selectable by the user and Back/Next buttons will remain the only navigation through them;
- Project Location label changed to Project Directory.
- Default project folder was changed from executable location to home() folder;
- Left margin was aligned with the right one;
- 'Project Type' became 'Type of Project';
- 'Add sources' became 'Add Design Files';
- 'Add constraints' became 'Add Constraint Files'
- Select Device -> Select Target Device.
image


src/NewProject/new_project_dialog.cpp
```diff
@@ -49,25 +89,29 @@ QString newProjectDialog::getProject() {
 
 void newProjectDialog::Reset() {
   m_index = INDEX_LOCATION;
-  for (int i = 0; i < ui->m_stackedWidget->count(); i++) {
-    auto w = ui->m_stackedWidget->widget(i);
-    ui->m_stackedWidget->removeWidget(w);
-    delete w;
-  }
+  ui->m_tabWidget->clear();
```
skyler-rs, line 92: This change will probably result in the tabs of m_tabWidget [not getting deleted ](https://doc.qt.io/qt-6/qtabwidget.html#clear)on reset, is that intentional?

skyler-rs (approved):
Looks good assuming the TabWidget clear() not deleting tabs is intended. 

Beyond that the UI looks good. I did try to click the tabs a few times w/o thinking so I'm pretty sure we'll get a request for that down the road, but we can save that for later once we've considered how to enable switching tabs when they might depend on previous data.

ravic-rs (changes requested):
Needs update on screens and text clean up.

src/NewProject/new_project_dialog.cpp
TarasSydor: That's intentional, yes - widgets are created with parents and will be deleted anyway. 
Still, it might be that after removing them from QTabWidget parent gets reset and we get a memory leak indeed. Will check that.

TarasSydor:
> Needs update on screens and text clean up.

Ravi, what do you mean? Should I add a screenshot of each changed tab to the PR description?
I added the list of all the updates. thought that adding a screenshot for each tab will make the description too big,

src/NewProject/new_project_dialog.cpp
TarasSydor: image

Confirmed - forms remain 'parented' after clearing the tab widget. They will be removed by QObject parent mechanism when the dialog gets destroyed - with current implementation at application exit.

KochynVolodymyr:
Did you fix layout? I don't see it on the screen?

src/NewProject/new_project_dialog.cpp
```diff
@@ -3,14 +3,55 @@
...
 #include "Compiler/Compiler.h"
 #include "MainWindow/Session.h"
 #include "ui_new_project_dialog.h"
+
 extern FOEDAG::Session *GlobalSession;
 using namespace FOEDAG;
 
+namespace {
+// Custom style to get horizontally shown tab labels in QTabWidget
+class CustomTabStyle : public QProxyStyle {
```
KochynVolodymyr, line 18: Could you please separate this in another file and put near FoedagStyle class.

src/NewProject/new_project_dialog.cpp
KochynVolodymyr: so technically this is no memory leak but it behaves like that. Every time we open this widget we use new memory and not free the one we don't need.

TarasSydor:
> Did you fix layout? I don't see it on the screen?

yes, I aligned left and right margins. The screenshot got slightly truncated on the right. attached a new one

ravic-rs (approved):
Some UI cleanup will be done in a separate PR.

PR #685, opened 2022-09-27
[GEMINIEDA-370] Extend set_top_module tcl command with optional -work parameter;

TarasSydor:
This pull request concludes GEMINIEDA-370 issue and fixes RTL management.

It makes sure libraries, added via -work tag are handled properly, based on language used.
>
- set_top_module got extended with optional -work parameter;
- '-L' tag is no longer supported as an input to add_design_file tcl command;
- ImportLibs field was removed from sources view; 
image
- -work libraries are passed to -import yosys directive based on language: top level library gets always passed, while others only for verilog files;
- help files adapted.

> - [+] Engine
> - [+] Documentation
> - [] Regression tests

Here are some screenshots demonstrating how tcl script change affects yosys output (1st screenshot is tcl script change, 2nd - yosys script change as a result):
image
image

image
image


TarasSydor:
the issue with duplicated 'topLib' to be fixed in a following commit.

src/ProjNavigator/tcl_command_integration.cpp
```diff
@@ -44,6 +44,7 @@ bool TclCommandIntegration::TclSetTopModule(int argc, const char *argv[],
   const QString topModule = QString(argv[1]);
   m_projManager->setCurrentFileSet(m_projManager->getDesignActiveFileSet());
   int ret = m_projManager->setTopModule(topModule);
+  if (argc == 4) m_projManager->setTopModuleLibrary(QString(argv[3]));
```
skyler-rs, line 47: Should this only fire when exactly 4 args are provided or is argc > 4 a possible valid scenario as well?

skyler-rs (approved):
Looks good as long as the argc == 4 portion mentioned above is intended.

src/ProjNavigator/tcl_command_integration.cpp
ravic-rs: Similar question. Do we have a argument parser to consume as needed instead of hard-coding numbers here ?

src/Compiler/Compiler.cpp
```diff
@@ -378,6 +374,9 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
     std::string actualType = "VERILOG_2001";
     Design::Language language = Design::Language::VERILOG_2001;
     auto file = StringUtils::toLower(argv[1]);
+    if (file == "-work")  // if optional -work parameter goes first, the first
```
KochynVolodymyr, line 377: very strange argument handling, would be good to improve this

src/Compiler/CompilerOpenFPGA.cpp
```diff
@@ -1030,18 +1031,23 @@ bool CompilerOpenFPGA::Synthesize() {
           break;
         case Design::Language::VERILOG_2001:
           lang = "-vlog2k";
+          importDesignFilesLibs = true;
```
KochynVolodymyr, line 1034: why not under BLIF/EBLIF case ? why not one line instead of 5?

src/ProjNavigator/tcl_command_integration.cpp
```diff
@@ -41,10 +41,28 @@ bool TclCommandIntegration::TclSetTopModule(int argc, const char *argv[],
     return false;
   }
 
-  const QString topModule = QString(argv[1]);
+  auto topModule = QString{};
```
KochynVolodymyr, line 44: this could be done more simple. Like:

  const QString topModule = QString(argv[1]);
  QString topModuleLib{};
  if (argc > 3) {
    if (QString{argv[2]} != "-work") {
      out << "Incorrect syntax .....";
      return false;
    }
    topModuleLib = QString(argv[3]);
  }
  m_projManager->setCurrentFileSet(m_projManager->getDesignActiveFileSet());
  int ret = m_projManager->setTopModule(topModule);
  m_projManager->setTopModuleLibrary(topModuleLib);
  if (0 == ret) update();
  return true;

src/Compiler/Compiler.cpp
TarasSydor: I moved that logic. now it's inside the loop and should be more readable

KochynVolodymyr:
Need approve from @alain-rs

src/Compiler/Compiler.cpp
```diff
@@ -368,34 +364,26 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
...
           "Incorrect syntax for add_design_file <file(s)> "
-          "[-work libraryName] [-L libraryName1 -L libraryName2]"
+          "[-work libraryName]"
           "<type (-VHDL_1987, -VHDL_1993, -VHDL_2000, -VHDL_2008 (.vhd "
           "default), -V_1995, "
           "-V_2001 (.v default), -SV_2005, -SV_2009, -SV_2012, -SV_2017 (.sv "
           "default))>");
       return TCL_ERROR;
     }
-    std::string actualType = "VERILOG_2001";
```
alaindargelas, line 378: If this is not set here anymore, where is the default language set ?

alaindargelas: Got it.

alaindargelas (approved):

ravic-rs (approved):
Looks good.

PR #687, opened 2022-09-28
[RG-94] Add language selection for project wizard

KochynVolodymyr:
Add language selection when user creates project. Blocked by https://github.com/os-fpga/FOEDAG/pull/685

 - [x] Library: newproject, projnavigator
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests

### screenshots
image
yosis script:
image

### DIFF
--- a/src/NewProject/source_grid.cpp
+++ b/src/NewProject/source_grid.cpp
@@ -1,6 +1,7 @@
 #include "source_grid.h"
 
 #include <QComboBox>
+#include <QDebug>
 #include <QDirIterator>
 #include <QFileDialog>
 #include <QHeaderView>
@@ -21,7 +22,8 @@ static const auto NAME_COL = QObject::tr("Name");
 static const auto LIBRARY_COL = QObject::tr("Library");
 static const auto LANGUAGE_COL = QObject::tr("Language");
 static const auto LOCATION_COL = QObject::tr("Location");
-static const int LANG_COL{4};
+static const int LIBRARY_COL_NUM{2};
+static const int LANG_COL_NUM{3};
 
 static const auto CONSTR_FILTER = QObject::tr("Constraint Files(*.sdc)");
 static const auto DESIGN_SOURCES_FILTER = QObject::tr(
@@ -141,6 +143,10 @@ QList<filedata> sourceGrid::getTableViewData() { return m_lisFileData; }
 
 void sourceGrid::currentFileSet(const QString &fileSet) {
   m_currentFileSet = fileSet;
+}
+
+void sourceGrid::selectRow(int row) {
+  if (row >= 0 && row < m_model->rowCount()) m_tableViewSrc->selectRow(row);
 }
 
 void sourceGrid::AddFiles() {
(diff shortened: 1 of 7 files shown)


src/NewProject/source_grid.cpp
```diff
@@ -364,6 +404,18 @@ void sourceGrid::onItemChanged(QStandardItem *item) {
     m_lisFileData[itemIndex.row()].m_importLibraries = item->text();
 }
 
+void sourceGrid::languageHasChanged() {
+  QComboBox *combo = qobject_cast<QComboBox *>(sender());
```
skyler-rs, line 408: Currently not a problem because we are connecting to a QComboBox, but to stay safe, probably best to check that the qobject_cast succeeded (combo isn't null) incase someone accidentally connects a non-qcombobox widget to this signal later on.

skyler-rs (approved):
Looks good once a null check is added to the qobject_cast result

src/NewProject/source_grid.cpp
KochynVolodymyr: good catch! thanks. Fixed

src/NewProject/source_grid.cpp
```diff
@@ -378,3 +439,30 @@ QStringList sourceGrid::GetAllDesignSourceExtentions() const {
...
+
+void sourceGrid::initLanguageCombo(int row, const QVariant &data) {
+  auto combo = CreateLanguageCombo();
+  combo->setCurrentIndex(combo->findData(data));
+  m_tableViewSrc->setIndexWidget(m_model->index(row, LANG_COL_NUM), combo);
+  connect(combo, SIGNAL(currentIndexChanged(int)), this,
+          SLOT(languageHasChanged()));
+}
+
+bool operator==(const FOEDAG::filedata &a, const FOEDAG::filedata &b) {
```
TarasSydor, line 451: Isn't it a default implementation? field by field comparison. why do we need its explicit implementation?

TarasSydor, line 451: Isn't it a default implementation? field by field comparison. why do we need its explicit implementation?

src/NewProject/source_grid.cpp
```diff
@@ -351,19 +381,50 @@ bool sourceGrid::IsFileDataExit(filedata fdata) {
   return false;
 }
 
+QComboBox *sourceGrid::CreateLanguageCombo() {
+  auto combo = new QComboBox;
+  combo->addItem("BLIF", Design::Language::BLIF);
```
TarasSydor, line 386: Just wondering - should the user be able to change the language of, for example, *.vhdl to VERILOG 2001?
if we want to forbid it then we may want to fill the combo based on automatically deduced language within 'sibling' languages. For *.sv files only SV 2005 - SV 2017 would be available then.

KochynVolodymyr: good point for backlog.

TarasSydor (approved):
looks good.

ravic-rs:
This might require support to tcl level, 
https://github.com/os-fpga/FOEDAG/pull/685/files#diff-55ad8fbd24c664c949cc02d0c581890f96e1da2d06272b1475ecbf39ceaf0bfdR434

KochynVolodymyr:
> This might require support to tcl level, https://github.com/os-fpga/FOEDAG/pull/685/files#diff-55ad8fbd24c664c949cc02d0c581890f96e1da2d06272b1475ecbf39ceaf0bfdR434

I didn't get it

ravic-rs:
> Collaborator

That was a note for me. I will let you know in morning.

ravic-rs (approved):
Requires a screenshot example of how this UI selection reflects at compiler level.

PR #689, opened 2022-09-29
check vpr version

alain-rs:


src/Compiler/CompilerOpenFPGA.cpp
```diff
@@ -1590,9 +1590,22 @@ bool CompilerOpenFPGA::Placement() {
     }
   }
 
+  // VPR Version checking, until full migration to version >= 8.0
+  std::string vprVersionCmd = m_vprExecutablePath.string() + " --version";
+  std::ostringstream ver;
+  FileUtils::ExecuteSystemCommand(vprVersionCmd, &ver);
+  bool version7 = true;
+  if (ver.str().find("Version: 8") != std::string::npos) {
```
KochynVolodymyr, line 1598: if version become 9 this condition fails. Is that accepted?

alaindargelas:
That code will be here for 1 month maximum. It is mentioned in the comment, it is a temporarily solution.

PR #691, opened 2022-09-30
[RG-99] Adding logic to find litex enabled python interpreter for IP Catalog

skyler-rs:
> - [x] To address an existing issue. If so, please provide a link to the issue: [RG-99]

> [FileUtils::LocateExecFile](https://github.com/os-fpga/FOEDAG/blob/f937d2954f15071d256b1f05019d1751864a3f17/src/Utils/FileUtils.cpp#L127) is used to search for a system copy of python3. LocateExecFile currently only works for unix. Additionally, this method of looking for a system copy of python3 will most likely fail to find the correct version of python where we've installed litex modules.
>
> A new function, FileUtils::FindFileRecursive() was added. This is used in conjunction with raptor changes to load a local copy of python that has litex/migen modules. This environment doesn't exist in foedag so in this case, we fallback to LocateExecFile for now.

> 
> - [x] Library: IPGenerate, Utils
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

> - [x] None

> ### Testing
> - This has been manually tested in Foedag and Raptor
> - This is a background path change for the most part and the path change is only observable in raptor where the python venv and litex modules are installed. Raptor doesn't appear to have unittest support now so i'd suggest we explore adding raptor side testing after code freeze.

### DIFF
--- a/src/IPGenerate/IPCatalog.cpp
+++ b/src/IPGenerate/IPCatalog.cpp
@@ -45,6 +45,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "Compiler/WorkerThread.h"
 #include "IPGenerate/IPCatalog.h"
 #include "MainWindow/Session.h"
+#include "Utils/FileUtils.h"
 #include "Utils/ProcessUtils.h"
 #include "Utils/StringUtils.h"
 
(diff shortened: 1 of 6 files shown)


ravic-rs:
image
Windows path
https://docs.python.org/3/using/windows.html

PR #697, opened 2022-10-03
[RG-107] Add additional fields to new project dialog

skyler-rs:
This is subtask 1 of 2 from RG-106, RG-108 will be addressed in a different PR.
> - [x] To address an existing issue. If so, please provide a link to the issue: rg-107

> New Project dialog provides a limited set of fields
>
> Include Path, Library Path, Library Ext, and Set Macro fields have been added
> image

> 
> - [x] Library: NewProject
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

> - [x] None

> ### Testing
> - This is a gui sided change and we don't have a harness to drive the gui
> - Manual testing and debug was done to ensure that values set in the UI were properly stored and accessible from the project manager.
> - It seems like there might be some unrelated bugs in the use of includePaths where the relative path might be incorrect. As such, the following test was done manually to ensure that these changes are working as expected:

### Testcase
These steps will show what a normal tcl script generates w/ these options
- Start raptor gui
- execute the following commands in the tcl console:
  - create_design scriptTest
  - target_device GEMINI
  - add_include_path ./includePath
  - add_library_path ./libraryPath
  - add_library_ext .s .sv
  - set_macro P=10 D=20
  - analyze
- exit the program
- cat ./scriptTest/scriptTest_analyzer.cmd
  
These steps recreate the same scenario via the gui:
- Start raptor gui
- Click File > New Project
- Set the project name to projectTest
- Click next twice so you are on the "Add Design Files" tab
- Set the following values:
image
- Click next until the final wizard page and then click finish
- Once the project is created, run the analyze step via `analyze` in the tcl console or double clicking it in the task tree
- cat <your_path_to_projectTest>/projectTest_analyzer.cmd

If both *_anyalyzer.cmd files are the same then the code is passing the variables as expected
Results:
image


KochynVolodymyr (approved):

ravic-rs (changes requested):
Waiting on additional sanity checks at upstream.

skyler-rs:
Converting this to a draft while investigating a possible issue

src/NewProject/ProjectManager/project_manager.cpp
```diff
@@ -62,6 +66,39 @@ void ProjectManager::CreateProject(const ProjectOptions& opt) {
...
   setTopModuleLibrary(opt.topModuleLib);
 
+  std::string delimiter = " ";
+  // Set Library Paths
+  if (!opt.libraryPathList.isEmpty()) {
+    std::vector<std::string> tokens;
+    StringUtils::tokenize(opt.libraryPathList.toStdString(), delimiter, tokens);
+    for (auto file : tokens) {
+      std::string cmd = "add_library_path " + file;
+      GlobalSession->CmdStack()->push_and_exec(new Command(cmd));
```
KochynVolodymyr, line 76: I'm not sure about this. This is not like it done for all other fields. But it make sense because you don't need to copy logic inside tcl commands.

skyler-rs: Yeah i wasn't sure either, the original commit on the pr does it the more standard way, but I noticed that the add_library_path tcl code has special path logic i hadn't capture so i tried invoking via tcl to get the same results. I used https://github.com/os-fpga/FOEDAG/blob/477b70d0d638ee0fb1b8a2c45ac08d950258dbae/src/MainWindow/main_window.cpp#L482 as an example I think this might actually be how we are supposed to be updating the _cmd.log as the user drives the UI, but that's a discussion for later.

KochynVolodymyr: yes, probably

ravic-rs (approved):
Looks good.

PR #707, opened 2022-10-05
[GEMINIEDA-432] Add logic to skip generating IPs that have already been generated

skyler-rs:
> - [x] To address an existing issue. If so, please provide a link to the issue: [GEMINIEDA-432]

> ipgenerate w/o parameters generates all ips
>
> - add List option to IPGenerateOpt to allow for passing of module names
> - added -modules option to ipgenerate which will only generate the passed IPs based off the moduleNames passed

> 
> - [x] Library: IPGenerate, Compile, IpConfigurator
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

> - [x] None

> ### Testing
> - A new test has been added to verify the new -modules argument added to ipgenerate
> - This was manually tested in raptor gui


src/IPGenerate/IPGenerator.cpp
```diff
@@ -464,6 +468,9 @@ bool IPGenerator::Generate() {
           m_compiler->ErrorMessage("IP Generate, " + help.str());
           return false;
         }
+
+        // Mark this instance as generated
+        inst->setGenerated(true);
```
alaindargelas, line 473: This is not correct, what if the parameters have changed, then it needs to be re-generated.
What you need is to have a Generate(ip_instance)  which takes a parameter to generate the IP under generation in the GUI.
And for the ipgenerate command, there is already (double check) a logic that geneates incrementaly the the parameters have not changed, if that is not working fix it.
But this high level status as you propose it is not correct.

alaindargelas (changes requested):

src/IPGenerate/IPGenerator.cpp
skyler-rs: Hey Alain, could you check this logic I added awhile back and see if it addresses your concern? https://github.com/skyler-rs/FOEDAG/blob/d886c3e2b9ae775c3c150381dd7ff86b41497823/src/IPGenerate/IPGenerator.cpp#L246
At some point i had to add logic for to replace an existing instance if a user reconfigured, w/ the same module name so any re-call off configure_ip will replace the existing instance which will then get generated on next ipgenerate and the instance will then be marked as generated.

If this approach is not appropriate, could you point me to the logic you mentioned? I don't see it in IPGenerator::Generate()

alaindargelas:
The logic is there previousbuffer == buffer
It reuses the IP on disk (Does not generate if the parameters are there).
So the boolean flag you added is not needed.

- your button in the GUI in the IP Catalog (IP Generate) when pressed, should only generate the IP in question (Pass an optional argument to the generate command).

- The IPGenerate action in the task list should generate all IPs (It does that today).

skyler-rs:
Gotcha that makes sense, thanks for clarifying. I will fix it first thing tomorrow.

skyler-rs:
Description has been updated to reflect the new code changes.

src/IPGenerate/IPGenerator.cpp
skyler-rs: I've refactored this PR to pass a module name as you suggested. I copied the SynthMoreOpts approach to pass the modules name(s) and added a List type to IPGenerateOpt to handle the scenario. Let me know if you were expected a different method for passing the module name.

alaindargelas (approved):

PR #729, opened 2022-10-12
[RG-135] Enable Stop button when either task manager or console is in progress

TarasSydor:
Given PR aims to fix RG-135 issue, which occurs at times while evaluating the project,

According to previous implementation, Stop action was only enabled when both, the console and the task manager, are in progress. Due to multiple status changes within the project compilation, sometimes console was no longer running while task manager still had in progress status. 

>
Stop button will be enabled whenever one of them is in progress. It makes sense anyway, as there is something we can stop.


ravic-rs (approved):
Looks good.

PR #732, opened 2022-10-12
Fixed compile order for design files

KochynVolodymyr:
Fixed design files compilation order.

 - [x] Library: newproject
 - [ ] Engine
 - [ ] Documentation
 - [x] Regression tests

 ### screenshots
**Expected:**
image

**Actual:**
image

### DIFF
--- a/src/NewProject/ProjectManager/project_manager.cpp
+++ b/src/NewProject/ProjectManager/project_manager.cpp
@@ -13,6 +13,7 @@
 #include "Compiler/CompilerDefines.h"
 #include "MainWindow/Session.h"
 #include "Utils/StringUtils.h"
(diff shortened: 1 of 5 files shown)


ravic-rs:
@KochynVolodymyr was it always broken ? I did notice this but think about the order then. 
Map and vector combo seems like the option since there's no option in std, atleast not that I'm aware off.

KochynVolodymyr:
> @KochynVolodymyr was it always broken ? I did notice this but didn't think about the order then. Map and vector combo seems like the option since there's no option in std, atleast not that I'm aware off.

@ravic-rs It appeared after we added Compile Unit

skyler-rs (approved):

PR #737, opened 2022-10-13
[GEMINIEDA-471] Add language mismatch check for compile units in new project / project settings

skyler-rs:
> - [x] To address an existing issue. If so, please provide a link to the issue: [GEMINIEDA-471]

> Error conditions were intentionally ignored
>
> We now check Compile Unit groups to ensure that each file has the same language type and alert the user with a message box if not:
image

> 
> - [x] Library: NewProject
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

> ### Testing
> - Manually tested in foedag and raptor for new projects and edited projects.

### DIFF
--- a/src/NewProject/new_project_dialog.cpp
+++ b/src/NewProject/new_project_dialog.cpp
@@ -191,7 +191,66 @@ void newProjectDialog::ResetToProjectSettings() {
   m_addSrcForm->SetBasePath(m_projectManager->getProjectPath());
 }
 
+// This will check the project wizard for potential conflicts
+// If an errStr is passed in, the found issues will be appended
+// to it.
+// This function is intended to grow as more potential conflicts
+// are discovered.
+bool newProjectDialog::ValuesValid(QString *errStr /*nullptr*/) {
+  QString temp;
+  if (!errStr) {
+    errStr = &temp;
+  }
+
+  // Check for Language type mistmatches in Compile Units
+  auto conflictKeys = FindCompileUnitConflicts();
+  if (!conflictKeys.isEmpty()) {
+    *errStr += "The following Compile Units have a language mismatch: " +
+               conflictKeys.join(", ") +
+               "\nReturn to the Design Files tab to fix this.";
+  }
+
+  return errStr->isEmpty();
+}
+
+QList<QString> newProjectDialog::FindCompileUnitConflicts() {
+  // Group files by Compile Unit (m_groupName)
+  QMap<QString, QList<filedata>> fileGroups{};
+  for (const filedata &fdata : m_addSrcForm->getFileData()) {
+    if (!fdata.m_groupName.isEmpty()) {
+      fileGroups[fdata.m_groupName].append(fdata);
+    }
+  }
+
+  // Determine if any group has a language mismatch
+  QStringList conflictKeys{};
+  // Step through keys
+  for (auto key : fileGroups.keys()) {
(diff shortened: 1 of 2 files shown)


src/NewProject/new_project_dialog.cpp
```diff
@@ -191,7 +191,66 @@ void newProjectDialog::ResetToProjectSettings() {
...
 
+// This will check the project wizard for potential conflicts
+// If an errStr is passed in, the found issues will be appended
+// to it.
+// This function is intended to grow as more potential conflicts
+// are discovered.
+bool newProjectDialog::ValuesValid(QString *errStr /*nullptr*/) {
+  QString temp;
+  if (!errStr) {
+    errStr = &temp;
```
KochynVolodymyr, line 202: Undefined behavior here

src/NewProject/new_project_dialog.h
```diff
@@ -75,6 +75,8 @@ class newProjectDialog : public QDialog {
   void UpdateDialogView(Mode mode = NewProject);
   void ResetToNewProject();
   void ResetToProjectSettings();
+  bool ValuesValid(QString* errStr = nullptr);
```
KochynVolodymyr, line 78: const would be good here

src/NewProject/new_project_dialog.h
```diff
@@ -75,6 +75,8 @@ class newProjectDialog : public QDialog {
...
+  QList<QString> FindCompileUnitConflicts();
```
KochynVolodymyr, line 79: const would be good here

src/NewProject/new_project_dialog.h
skyler-rs: Fixed, thanks for the reminder, i'll eventually remember to check for this ;)

src/NewProject/new_project_dialog.cpp
KochynVolodymyr: I don't see fix of this problem. This is critical.

skyler-rs: Sorry, wrote a comment and hit resolve and forgot that it doesn't submit the comment like that. This is just to allow the user to ignore the errStr if they want. If they don't pass one, then we just point at a temp QString that we won't use (the errors will get written to avoid adding logic to every errStr +=, but the message itself won't be used besides for determining the bool status at the end). Let me know if i'm missing your point though.

KochynVolodymyr: yes, you are right. But this looks like code smell. It is better to do like
if (errStr) *errStr = ...

skyler-rs: Wouldn't you to new and delete a QString then since errStr is a pointer?

KochynVolodymyr: We don't need it. If pointer is not nullptr we do something if it is nullptr we do nothing. 
But there is better solution like: 
std::pair<bool, QString> ValuesValid() const;
and you don't need to care about pointers

KochynVolodymyr (approved):

src/NewProject/new_project_dialog.cpp
skyler-rs: Forgot to hit comment again before resolve ;p
Sounds good, i had considered pair at first, but thought it might be overkill. I've refactored the code to use std::pair now.

PR #739, opened 2022-10-14
[GEMINIEDA-483] Adding logic to print topModule after analyze

skyler-rs:
> - [x] To address an existing issue. If so, please provide a link to the issue: [GEMINIEDA-483]

> top module info isn't printed, but the info can be found in <project_dir>/port_info.json after an analyze step
>
> The analyze step has been updated to read the "topModule" property from port_info.json and print it at the end of the task
> The message will print after a new run: 
image

> and when the design has already been analyzed:
image

> 
> - [x] Library: Compiler
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

> ### Testing
> - Manually tested in raptor. Foedag will not demonstrate this functionality as its dummy compiler doesn't generate port_info.json. As such, the logic to print topModule does not exist in the dummy compiler.

### DIFF
--- a/src/Compiler/CompilerOpenFPGA.cpp
+++ b/src/Compiler/CompilerOpenFPGA.cpp
@@ -732,8 +732,8 @@ bool CompilerOpenFPGA::DesignChanged(
 }
 
 bool CompilerOpenFPGA::Analyze() {
-  auto printTopModule = [](const std::filesystem::path& filePath,
-                           std::ostream* out) {
+  auto printTopModules = [](const std::filesystem::path& filePath,
+                            std::ostream* out) {
     // Check for "topModule" in a given json filePath
     // Assumed json format is [ { "topModule" : "some_value"} ]
     if (out) {
@@ -741,11 +741,15 @@ bool CompilerOpenFPGA::Analyze() {
         std::ifstream file(filePath);
         json data = json::parse(file);
         if (data.is_array()) {
-          data = data.at(0);
-          if (data.contains("topModule")) {
-            std::string val = data.value("topModule", "");
-            (*out) << "Top Module: " << val << std::endl;
-          }
+          std::vector<std::string> topModules;
+          std::transform(data.begin(), data.end(),
+                         std::back_inserter(topModules),
+                         [](json val) -> std::string {
+                           return val.value("topModule", "");
+                         });
+
+          (*out) << "Top Modules: " << StringUtils::join(topModules, ", ")
+                 << std::endl;
         }
       }
     }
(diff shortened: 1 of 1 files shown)


skyler-rs:
@alain-rs One note, i used the port_info.json from AES_DECRYPT to determine what the json format was. From that i assumed that it's just one object in an array that has "topModule" defined. If there is a scenario where there could be multiple objects in that array that all have a "topModule" defined, let me know and I'll see about updating the logic.

alaindargelas:
There could be more than 1 top level modules. That is the purpose of this code, to print all the top level modules.
The json format does support any number of top level modules.

skyler-rs:
Gotcha, i'll update the logic to iterate over the array, thanks!

alaindargelas:
Users will subsequently select one of the top level modules and enter it in the project wizard "top module" field  if they have not done so yet.

src/Compiler/CompilerOpenFPGA.cpp
```diff
@@ -729,6 +732,29 @@ bool CompilerOpenFPGA::DesignChanged(
...
+    if (out) {
+      if (FileUtils::FileExists(filePath)) {
+        std::ifstream file(filePath);
+        json data = json::parse(file);
+        if (data.is_array()) {
+          std::vector<std::string> topModules;
+          std::transform(data.begin(), data.end(),
+                         std::back_inserter(topModules),
+                         [](json val) -> std::string {
+                           return val.value("topModule", "");
```
ravic-rs, line 748: In case of multiple top modules, we still have one key with multiple values, correct?

ravic-rs (approved):
Looks good, 1 comment to address.

src/Compiler/CompilerOpenFPGA.cpp
skyler-rs: No that's not how I understand it from Alain's reply. I originally was reading the json assuming there was only 1 element in the array. Now the code will step through multiple objects in the top array of the json and grab "topModule" from each object. If that's not the case and there is only 1 object in an unnecessary array and "topModule" contains space separated modules, that should still work as my code will read the first array value which in that case would just have multiple stored in it. Only difference in that scenario is the values would display space separated instead of comma separated which is fine since this is visual, user level info that doesn't require a parseable formatting.

alaindargelas:
Use this testcase to test for multi top level module:

File dut.sv:

module top1();
endmodule

module top2();
endmodule

skyler-rs:
> Use this testcase to test for multi top level module:
> 
> File dut.sv:
> 
> module top1(); endmodule
> 
> module top2(); endmodule

Thanks for the testcase. Yep, looks like it puts two objects in an array which is what i refactored for so we are good.  
**ports_info.json:**
```
[
    {
        "topModule": "top1"
    },
    {
        "topModule": "top2"
    }
]
```

ravic-rs:
> > Use this testcase to test for multi top level module:
> > File dut.sv:
> > module top1(); endmodule
> > module top2(); endmodule
> 
> Thanks for the testcase. Yep, looks like it puts two objects in an array which is what i refactored for so we are good. **ports_info.json:**
> 
> ```
> [
>     {
>         "topModule": "top1"
>     },
>     {
>         "topModule": "top2"
>     }
> ]
> ```
Great! Will merge.

PR #746, opened 2022-10-18
Update pin package table

KochynVolodymyr:
Updates for pin planner
* load data from *.sdc file
* Add filter to combo boxes
* Update loaders for new csv table.
* Fixed couple issues related to sync between table and internal data.

- [x] Library: pinassignment
- [ ] Engine
- [ ] Documentation
- [x] Regression tests

### Screenshots
image
image

### DIFF
--- a/src/PinAssignment/PackagePinsLoader.cpp
+++ b/src/PinAssignment/PackagePinsLoader.cpp
@@ -21,6 +21,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "PackagePinsLoader.h"
 
 #include <QFile>
+#include <QSet>
 
 #include "nlohmann_json/json.hpp"
 using json = nlohmann::ordered_json;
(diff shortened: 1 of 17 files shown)


src/PinAssignment/PackagePinsLoader.cpp
```diff
@@ -43,23 +44,24 @@ std::pair<bool, QString> PackagePinsLoader::load(const QString &fileName) {
 #endif
   lines.pop_front();  // header
   PackagePinGroup group{};
+  QSet<QString> uniquePins;
   for (const auto &line : lines) {
```
skyler-rs, line 48: Not necessary for this check in, but code blocks like these could benefit from comments. The interaction between the different model, groups, etc isn't obvious if you aren't already familiar with the code.

skyler-rs (approved):

src/PinAssignment/PackagePinsLoader.cpp
KochynVolodymyr: agree

PR #750, opened 2022-10-19
[RG-139] Pin planner update, add Mode column

KochynVolodymyr:
Add mode column for Package pin table in pin planner

- [x] Library: pinassignment
- [ ] Engine
- [ ] Documentation
- [x] Regression tests

### Screenshots
image

### DIFF
--- a/etc/package_pin_info.json
+++ b/etc/package_pin_info.json
@@ -20,75 +20,81 @@
       "visible" : true
     },
     {
+      "name" : "Mode",
+      "description" : "Mode of the gearbox",
+      "colmumnid" : 3,
+      "visible" : true
+    },
+    {
       "name" : "Ref clock",
       "description" : "Clock name referenced by pin",
-      "colmumnid" : 3,
+      "colmumnid" : 4,
       "visible" : true
     },
     {
       "name" : "Bank",
       "description" : "IO bank name",
-      "colmumnid" : 4,
+      "colmumnid" : 5,
       "visible" : true
     },
     {
       "name" : "ALT",
       "description" : "Alternative function of pin",
-      "colmumnid" : 5,
+      "colmumnid" : 6,
       "visible" : true
     },
     {
       "name" : "Debug mode",
       "description" : "Available debug mode of pin",
-      "colmumnid" : 6,
+      "colmumnid" : 7,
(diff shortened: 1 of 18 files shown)


KochynVolodymyr:
@taoli4rs please take a look at the screenshot. This is how it looks like.

skyler-rs (approved):
Looks good, only thing that looks a bit odd is the hardcoding of the column number in the json (just prone to update mistakes if anything is added etc).

Since you're reading the json using an ordered format (`using json = nlohmann::ordered_json`), you could probably just increment columnId on each read and make it dynamic. 

That said, i can see how including it might provide a bit more portability incase some other system read the json in and wasn't doing it ordered.

No changes needed, just something to consider next time you're working with this code.

KochynVolodymyr:
> Looks good, only thing that looks a bit odd is the hardcoding of the column number in the json (just prone to update mistakes if anything is added etc).
> 
> Since you're reading the json using an ordered format (`using json = nlohmann::ordered_json`), you could probably just increment columnId on each read and make it dynamic.
> 
> That said, i can see how including it might provide a bit more portability incase some other system read the json in and wasn't doing it ordered.
> 
> No changes needed, just something to consider next time you're working with this code.

I don't want rely on reading order. It is less obvious.

taoli4rs:
> 
Layout LGTM. One question: when list of available modes for user to select, do we consider the port direction? Here in the screen shot, if design port "ct_vld" is input, then we should list modes of "Mode_XXXX_RX" only; similarly, if "ct_vld" is output, then we should list modes of "Mode_XXXX_TX" only.

KochynVolodymyr:
> @volodymyrkochyn > Layout LGTM. One question: when list of available modes for user to select, do we consider the port direction? Here in the screen shot, if design port "ct_vld" is input, then we should list modes of "Mode_XXXX_RX" only; similarly, if "ct_vld" is output, then we should list modes of "Mode_XXXX_TX" only.

I'm on it.

KochynVolodymyr:
@ravic-rs I guess this is ready

ravic-rs (approved):
Looks good.

taoli4rs:
LGTM. We can do more enhancement later.

PR #753, opened 2022-10-24
[RG-137]: Play loading animation while compilation in tasks table view

TarasSydor:
Given PR fulfills RG-137 issue and replaced Loading icon with an animation.

Qt can automatically show icons in table view, but Gif animations aren't supported. To play it, I had to use QLabel as the widget for Status column, as set animation to it on each paint.
The class to play gif animation is QMovie.
>
Status column now shows loading animation instead of static icon while compiling the project:
image
image

> 
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests


src/Compiler/TaskModel.cpp
```diff
@@ -89,7 +89,7 @@ QVariant TaskModel::data(const QModelIndex &index, int role) const {
         case TaskStatus::Fail:
           return QIcon(":/failed.png");
         case TaskStatus::InProgress:
-          return QIcon(":/loading.png");
+          return QString(":/loading.gif");
```
KochynVolodymyr, line 92: it doesn't make any sense. What is just return bool, e.g. true? or just 0 or 1 ?

ravic-rs: @TarasSydor ?

TarasSydor: fixed

KochynVolodymyr (approved):

ravic-rs (approved):
Looks good.

PR #758, opened 2022-10-25
Initial TCL testing framework prototype

skyler-rs:
Note: This PR isn't intended for merging.

This is to show a tcl/qt testing prototype. The goal is to show an initial idea on how we might automate our gui testing and seek feedback and possible issues etc.

More details on the implementation, future goals, and current issue can be found under [RG-144]

There are 3 basic examples under tests/TestGui/Automated to show what will be involved to automate our gui. Anything that uses a modal dialog requires some rather messy logic currently.


skyler-rs:
I've added a proof of concept way of manipulating the ui after a modal dialog starts blocking. This can be demonstrated in the console with `qt_dlg_test`. Next step is to come up with a queue of some sort where we can collect commands we'll fire while the modal dialog is blocking.

KochynVolodymyr:
Looks good to me. How to know that test passed? How to get actual and expected values and compare them? could you please provide something?

skyler-rs:
Added example tests to tests/TestGui/Automated. One demonstrates how to deal w/ a modal dialog, then other explores how we currently need to try/catch our scripts to catch errors when running w/ --script.

tests/TestGui/tempTesting.tcl
```diff
@@ -0,0 +1,36 @@
...
+# Note: Currently, only non-modal dialogs work, the following line only passes if the associated dlg has been changed to non-mmodal (from ->exec() to ->show())
+set dlg [waitForWidget -type dialog -objectName tasksDlg_Synthesis__SettingsDialog]
+
+# Generic widgets like an OK button can use the -parent argument to limit the search scope
+set ok [waitForWidget -parent $dlg -type button -text ok]
+qt_clickButton $ok
+
+# Not everything waits for widgets currently and you might run into a scenario where you need so delay for the test to pass
+# adding sleep X gives the ui a bit more time to finish its last operation before executing the next steps
+sleep 1000
```
ravic-rs, line 31: Can you expand of which "Not everything waits.."? The hardcoded sleep here seems to be dependent on the UI operation under test.

skyler-rs: Note: i've deleted the file in question, but [these lines](https://github.com/os-fpga/FOEDAG/pull/758/files#diff-12881b456815f16f4bb55ae85c2f2f00eb1dbe2182d4c73952445f656f448de2R27-R28) demonstrates the same issue asked about.

@ravic-rs when firing sequential UI changing commands that don't need to find a widget first, the system is executing those commands faster than the ui can update so we add a delay to let things catch up. We can probably remove these sleeps by writing additional waitForWidget commands to verify when a widget we expect is shown, but that requires additional development of the harness to target the desired widgets etc. Since the harness is a very minimal prototype right now, sleep 1000 just helps demonstrate the functionality w/o building out everything.

PR #761, opened 2022-10-26
[RG-142] Fixed reset compile steps

KochynVolodymyr:
Fixed reset compile steps.

- [x] Library: compiler
- [ ] Engine
- [ ] Documentation
- [ ] Regression tests

### Screenshot
image

### DIFF
--- a/src/Compiler/TaskManager.cpp
+++ b/src/Compiler/TaskManager.cpp
@@ -100,32 +100,25 @@ TaskManager::TaskManager(QObject *parent) : QObject{parent} {
             &TaskManager::taskStateChanged);
     connect((*task), &Task::finished, this, &TaskManager::runNext);
   }
-  QVector<Task *> tmp = {m_tasks[BITSTREAM]};
-  m_rollBack.insert(m_tasks[BITSTREAM_CLEAN], tmp);
-
-  tmp += m_tasks[POWER];
-  m_rollBack.insert(m_tasks[POWER_CLEAN], tmp);
-
-  tmp += m_tasks[TIMING_SIGN_OFF];
-  m_rollBack.insert(m_tasks[TIMING_SIGN_OFF_CLEAN], tmp);
-
-  tmp += m_tasks[ROUTING];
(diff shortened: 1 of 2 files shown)


src/Compiler/TaskManager.h
```diff
@@ -111,12 +111,12 @@ class TaskManager : public QObject {
...
   void run();
   void reset();
-  void rollBack(Task *t);
+  void cleanDownStreamStatus(Task *t);
 
  private:
   QMap<uint, Task *> m_tasks;
   QVector<Task *> m_runStack;
-  QMap<Task *, QVector<Task *>> m_rollBack;
+  QVector<Task *> m_taskQueue;
```
ravic-rs, line 119: What's the motivation to replace map with a vector ?

src/Compiler/TaskManager.cpp
```diff
@@ -228,10 +221,17 @@ void TaskManager::reset() {
...
 }
 
-void TaskManager::rollBack(Task *t) {
-  const auto &roll_back = m_rollBack.value(t);
-  for (const auto &task : roll_back) {
-    task->setStatus(TaskStatus::None);
+void TaskManager::cleanDownStreamStatus(Task *t) {
+  for (auto it{m_taskQueue.begin()}; it != m_taskQueue.end(); ++it) {
+    if (*it == t) {
+      // In case clean action, clean parent is required.
```
ravic-rs, line 227: What's the behavior when we don't have clean action for a task ?

ravic-rs (commented):
Two comments to address

src/Compiler/TaskManager.cpp
KochynVolodymyr: nothing happened in this case

src/Compiler/TaskManager.h
KochynVolodymyr: I make it more simple and more general. No need a map anymore

TarasSydor (approved):

PR #763, opened 2022-10-27
[RG-146] --project runs the project. run_project tcl command introduced

TarasSydor:
The goal of this PR is fixing RG-146 ticked.

After the OpenProject call the code also does start(), which runs opened project.
Next to that, run_project TCL command was registered, that also opens the project and immediately runs it.
>
open_project keeps on existing. run_project can be used in order to open and run the project immediately.
Run is also used in tests now.

> 
> - [ ] Engine
> - [ ] Documentation
> - [x] Regression tests


KochynVolodymyr (approved):

tests/TestGui/run_project_test/run_project.tcl
```diff
@@ -17,5 +17,5 @@
 #You should have received a copy of the GNU General Public License
 #along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-puts "OPENING PROJECT"; flush stdout ; open_project test.ospr
-puts "PROJECT OPENED"; flush stdout ; exit
+puts "RUNNING PROJECT"; flush stdout ; run_project test.ospr
```
ravic-rs, line 20: Error out at failure.
Refer: https://github.com/os-fpga/FOEDAG/blob/main/tests/TestBatch/test_ip_generate.tcl#L42

TarasSydor:
While working on running project validation I noticed incorrect behavior when using --replay. --script works as expected
latest commit fixes that. project run validation should still be implemented - appeared to be more complex than expected

ravic-rs:
> While working on running project validation I noticed incorrect behavior when using --replay. --script works as expected latest commit fixes that. project run validation should still be implemented - appeared to be more complex than expected

What was the incorrect behavior found?

TarasSydor:
> > While working on running project validation I noticed incorrect behavior when using --replay. --script works as expected latest commit fixes that. project run validation should still be implemented - appeared to be more complex than expected
> 
> What was the incorrect behavior found?

Project file wasn't recognized. That was caused by --script specific paths handling.
now fixed

tests/TestGui/run_project_test/run_project.tcl
TarasSydor: I did it in a bit another way but the idea is similar: if foedag.log contains 'failed', we exit with an error. output PASSED otherwise:
image

TarasSydor:
@ravic-rs  Here is the list of changes after the last commit:
- run_project now waits till run is finished;
- it checks tasks statuses. in case some failed - error will be reported;
- Makefile verifies whether foedag.log has any errors and interrupts the test.

src/MainWindow/main_window.cpp
```diff
@@ -269,11 +269,10 @@ void MainWindow::newDesignCreated(const QString& design) {
...
 
 void MainWindow::startStopButtonsState() {
-  const bool inProgress = m_taskManager->status() == TaskStatus::InProgress;
-  const bool consoleInProgress = m_console->isRunning();
-  startAction->setEnabled(!inProgress && !consoleInProgress);
+  startAction->setEnabled(m_taskManager->status() != TaskStatus::InProgress &&
+                          !m_console->isRunning());
   // Enable Stop action when there is something to stop
-  stopAction->setEnabled(inProgress || consoleInProgress);
+  stopAction->setEnabled(isRunning());
```
ravic-rs, line 275: Why are we changing the stopAction state here ? Seems like a refactoring, https://github.com/os-fpga/FOEDAG/pull/763/files#diff-9e31125b4e2b6bd449e09a6b478450cb54bf4fb9ddb23baadb06ad5d9a9bb149R315
correct ?

ravic-rs (changes requested):
1 comment to address.

ravic-rs:
> @ravic-rs Here is the list of changes after the last commit:
> 
> * run_project now waits till run is finished;
> * it checks tasks statuses. in case some failed - error will be reported;
> * Makefile verifies whether foedag.log has any errors and interrupts the test.

Can we add a testcase for failed error?

tests/TestGui/run_project_test/run_project.tcl
ravic-rs: We discussed about adding it the way test_ip_generate.tcl was implemented, hence the comment pointing to it. Why can't we use that approach? The grep way to check is external to the tcl. I would like to have the check built as part of the test tcl itself.

src/MainWindow/main_window.cpp
TarasSydor: We aren't changing it. The logic was just moved to another place. here is the implementation:
image
we use the same condition to make sure run_project waits till project finishes running. to avoid code duplication the logic was just moved to a separate method.

tests/TestGui/run_project_test/run_project.tcl
TarasSydor: fixed. verification was moved from Makefile to the tcl script

src/Compiler/Compiler.cpp
```diff
@@ -282,6 +286,53 @@ bool Compiler::BuildLiteXIPCatalog(std::filesystem::path litexPath) {
...
+      QCoreApplication::processEvents(QEventLoop::WaitForMoreEvents);
+  }
+  auto allTasks = compiler->GetTaskManager()->tasks();
+  for (auto task : allTasks) {
+    if (task && task->status() == TaskStatus::Fail) {
+      compiler->ErrorMessage(task->title().toStdString() + "task failed!");
+      return TCL_ERROR;
+    }
+  }
+  compiler->Message("Project run successful!");
```
ravic-rs, line 332: Please remove "!"

ravic-rs (approved):

ravic-rs:
@TarasSydor your forked branch is out of sync with "main" here. Please keep this in mind for long running PRs.

PR #764, opened 2022-10-28
Add logic to delete cached json when deleting ip

skyler-rs:
> - [x] Fixing an oversite in previous implementation

> - When deleting an ip from the UI w/ r-click > Delete IP, the ip build files were removed, but the cached .json file wasn't which would result in a failure to generate the actual ip files the next time someone tried to generate that IP again.
>
> - Abstracted the logic for determining the cached json path
> - Added logic to delete the cached json file when an IP is requested to be deleted from the UI.

> 
> - [x] Library: IPGenerate
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

> - [x] None

> ### Testing
> - Manually tested in Foedag and Raptor

### DIFF
--- a/src/IPGenerate/IPGenerator.cpp
+++ b/src/IPGenerate/IPGenerator.cpp
@@ -518,7 +518,7 @@ std::filesystem::path IPGenerator::GetCachePath(IPInstance* instance) const {
 
   ProjectManager* projManager = nullptr;
   if (m_compiler && (projManager = m_compiler->ProjManager())) {
-    std::string projectPath = m_compiler->ProjManager()->projectPath();
+    std::string projectPath = projManager->projectPath();
     auto def = instance->Definition();
     std::string ip_config_file =
         def->Name() + "_" + instance->ModuleName() + ".json";
--- a/src/IPGenerate/IPGenerator.cpp
+++ b/src/IPGenerate/IPGenerator.cpp
@@ -360,6 +360,11 @@ void IPGenerator::DeleteIPInstance(IPInstance* instance) {
       FileUtils::FileIsDirectory(buildPath)) {
     std::filesystem::remove_all(buildPath);
   }
+  // Delete the cached json file
+  auto filePath = GetCachePath(instance);
+  if (FileUtils::FileExists(filePath)) {
+    std::filesystem::remove_all(filePath);
+  }
 
   RemoveIPInstance(instance);
 }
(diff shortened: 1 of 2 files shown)


src/IPGenerate/IPGenerator.cpp
```diff
@@ -436,7 +437,8 @@ bool IPGenerator::Generate() {
         jsonF << "   \"build_name\": " << inst->OutputFile().filename() << ","
               << std::endl;
         jsonF << "   \"build\": true," << std::endl;
-        jsonF << "   \"json\": \"" << ip_config_file << "\"," << std::endl;
+        jsonF << "   \"json\": \"" << jasonfile.filename().string() << "\","
```
ravic-rs, line 440: Typo
```suggestion
        jsonF << "   \"json\": \"" << jsonfile.filename().string() << "\","
```

skyler-rs: Yeah that's a holdout from the original implementation, it's used throughout the function, should i rename it?

ravic-rs (commented):
1 comment to address

src/IPGenerate/IPGenerator.cpp
ravic-rs: Yes, it will help during the search for json.

skyler-rs: Fixed

src/IPGenerate/IPGenerator.cpp
```diff
@@ -510,3 +512,21 @@ std::filesystem::path IPGenerator::GetBuildDir(IPInstance* instance) const {
...
   return dir;
 }
+
+// This will return the path to this instance's cached json file
+std::filesystem::path IPGenerator::GetCachePath(IPInstance* instance) const {
+  std::filesystem::path dir{};
+
+  Compiler* compiler = GlobalSession->GetCompiler();
+
+  ProjectManager* projManager = nullptr;
```
KochynVolodymyr, line 522: I don't thing you need this.

skyler-rs: I was getting a compiler error when doing just GetCompiler() for some reason, the const qualifier was my only guess.

KochynVolodymyr: if (compiler && compiler->ProjManager()) should work. If not, please fix

skyler-rs: Oh i see what you mean now, i can just use m_compiler w/o a getter, i didn't realize the parent class already had its own ptr. Good catch! Fixed.

ravic-rs (approved):
Looks good. Fix to regenerate the ip after the user deletes (from the disk) it from the project.

PR #766, opened 2022-10-29
Support multiple sdc files

alain-rs:


alain-rs:
Prevent reset of Constraint object in between files.
Parse constraint files only once when added, and again when synthesis is ran

src/Compiler/Compiler.cpp
```diff
@@ -700,6 +708,12 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
         compiler->ErrorMessage(out.str());
         return TCL_ERROR;
       }
+    } else {
```
KochynVolodymyr, line 711: this is not working. why it was changed?

alaindargelas: Constraint file was being sourced twice. One in this method, another time in the project method. 
Especially obvious when multiple constraint files are used.
What is not working exactly? It is working fine with all the tests I did with multiple constraint files.

KochynVolodymyr: This code is not reachable since compiler->m_tclCmdIntegration always true

alaindargelas:
Then just remove this redundant code. It works with the other code path since you read_sdc in the other function.

PR #776, opened 2022-11-01
[RG-149] Enable multi-select in source tree

skyler-rs:
> - [x] To address an existing issue. If so, please provide a link to the issue: RG-149

> - Source tree only supports single selection
>
> - Source tree selection is changed to [ExtendedSelection](https://doc.qt.io/qt-6/qabstractitemview.html#SelectionMode-enum)
> - Open File, Remove File, Remove/Delete IP options have been updated to support multi select and only interact with related selections. See RG-149 for details.
> - Remove File confirmation message has been updated to list multiple files on multi select or the old message on single file selection
> image  
>
> Note how you can select anything in the tree, but when doing remove file, it only offers to remove the related source files. (In this image, SubWord.sv was r-clicked. Had an ip instance been r-clicked and Remove/Delete IP been clicked, then only the IP related instances would be interacted with as the other selections aren't related to IP).

> 
> - [x] Library: ProjNavigator
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

> ### Testing
> - Manually tested in Foedag and Raptor


src/ProjNavigator/sources_form.cpp
```diff
@@ -290,23 +294,55 @@ void SourcesForm::SlotRemoveFileSet() {
...
+      // Multi File Message
+      questionStr = tr("Are you sure you want to remove the following files "
+                       "from the project?\n %1\n\nThey "
+                       "will not be removed from the disk.")
+                        .arg(files.join("\n "));
+    }
     if (QMessageBox::question(this, {}, questionStr) == QMessageBox::No) return;
-    m_projManager->setCurrentFileSet(fileSet.toString());
-    int ret = m_projManager->deleteFile(strFileName);
-    if (0 == ret) {
```
skyler-rs, line 306: @volodymyrkochyn When switching to a loop to delete multiple files, I decided to remove the logic of not updating/saving on ret != 0 because with multiple files, the next successful delete would update/save anyways. Prior to this change, i believe the same could be achieved by deleting a file (let's assume it somehow returned !0 thus not saving) and then deleting a different file that does return 0 as expected, in that case the project will still save even though the logic choose to not save on the first file failure so I believe the changed logic is still the same basically.

Let me know if you see any issue w/ this or if we really need to error out when the project manager fails to delete a file.

src/ProjNavigator/sources_form.h
```diff
@@ -111,6 +111,7 @@ class SourcesForm : public QWidget {
   static QString StripPath(const QString& path);
   void showAddFileDialog(GridType gridType);
   void AddIpInstanceTree(QTreeWidgetItem* topItem);
+  QStringList SelectedIpModules();
```
KochynVolodymyr, line 114: const is missing

src/ProjNavigator/sources_form.cpp
```diff
@@ -264,15 +264,19 @@ void SourcesForm::SlotAddFile() {
...
-  if (item == nullptr) {
-    return;
+  // Get selected file names
+  for (auto item : m_treeSrcHierachy->selectedItems()) {
+    // Using WhatsThis role to ensure selection is a file
+    // OpenFile quietly aborts on bad filenames, so this check could be removed
+    // if a desired type is being missing
+    if (item->data(0, Qt::WhatsThisPropertyRole)
+            .toString()
+            .contains("fileitem")) {
```
KochynVolodymyr, line 274: I don't see using "fileitem" anywhere. Where is it come from?

src/ProjNavigator/sources_form.cpp
```diff
@@ -290,23 +294,55 @@ void SourcesForm::SlotRemoveFileSet() {
...
+    }
     if (QMessageBox::question(this, {}, questionStr) == QMessageBox::No) return;
-    m_projManager->setCurrentFileSet(fileSet.toString());
-    int ret = m_projManager->deleteFile(strFileName);
-    if (0 == ret) {
-      UpdateSrcHierachyTree();
-      m_projManager->FinishedProject();
+
+    // Loop through and remove files
+    for (QPair<QString, QVariant> selection : selections) {
```
KochynVolodymyr, line 340: why not const reference here?

src/ProjNavigator/sources_form.cpp
KochynVolodymyr: I'm fine with it.

src/ProjNavigator/sources_form.cpp
```diff
@@ -290,23 +294,55 @@ void SourcesForm::SlotRemoveFileSet() {
...
+  // Use the r-clicked item for determining what item type we are working with
+  QTreeWidgetItem *refItem = m_treeSrcHierachy->currentItem();
+  auto itemType = refItem->data(0, Qt::WhatsThisPropertyRole);
+
+  // Bail on no selection
+  if (refItem == nullptr) return;
+  if (!selectedItems.count()) return;
+
+  // Track string names for confirmation dialog
+  QStringList files;
```
KochynVolodymyr, line 308: this is duplication of 'selections'. Same data with same condition goes to both containers. Right?

src/ProjNavigator/sources_form.cpp
skyler-rs: "fileitem" is part of a string used in the WhatsThisPropertyRole for each file type. You can find the defines here https://github.com/os-fpga/FOEDAG/blob/6b95d2bb9dacd37585c0659e03c5c7c7827bf9bc/src/ProjNavigator/sources_form.h#L16. So we are basically targeting SRC_TREE_DESIGN_FILE_ITEM, SRC_TREE_CONSTR_FILE_ITEM, SRC_TREE_SIM_FILE_ITEM & SRC_TREE_IP_FILE_ITEM with this.

src/ProjNavigator/sources_form.cpp
skyler-rs: Yeah this is just to simplify the creation of the confirmation message since i'd need a way to split out a stringlist (so i can use join) later anyways.

src/ProjNavigator/sources_form.cpp
skyler-rs: Didn't think of it, fixed

src/ProjNavigator/sources_form.h
skyler-rs: Fixed

ravic-rs (approved):

PR #779, opened 2022-11-02
[RG-152] Enable menu bar in welcome page. show File and Help menus

TarasSydor:
Given PR fulfills RG-152 ticket and enables menu bar when welcome page is shown.

The idea is to 'mark' menu actions, that should be shown at the welcome page, with a special property.
When showing or hiding welcome page, we loop over all menus and their children and check that property. If it's set and welcome page is shown - action becomes visible. 

Two states of visibility were introduced: Fully and Partially visible. It can be set for parent actions, which will show all the sub items on the welcome page (e.g. Help menu or Recent projects ).

Now Welcome Page looks as follows:
image
image


skyler-rs (approved):
Implementation looks great, this logic will help protect against new actions accidentally getting added to the WelcomePage menus.

src/MainWindow/main_window.cpp
```diff
@@ -513,10 +530,30 @@ void MainWindow::createToolBars() {
...
-  debugToolBar->setVisible(show);
+void MainWindow::updateMenusVisibility(const bool welcomePageShown) {
+  updateActionsVisibility(menuBar()->actions(), welcomePageShown);
+
+  fileToolBar->setVisible(!welcomePageShown);
+  debugToolBar->setVisible(!welcomePageShown);
+}
+
+void MainWindow::updateActionsVisibility(const QList<QAction*>& actions,
+                                         const bool welcomePageShown) {
```
KochynVolodymyr, line 541: you don't need const here

src/MainWindow/main_window.cpp
```diff
@@ -54,6 +54,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "Utils/FileUtils.h"
 #include "foedag_version.h"
 
+#define WELCOME_PAGE_MENU_PROP "showOnWelcomePage"
```
KochynVolodymyr, line 57: could you please replace #define with const QString {};?

src/MainWindow/main_window.cpp
```diff
@@ -513,10 +530,30 @@ void MainWindow::createToolBars() {
   debugToolBar->addAction(stopAction);
 }
 
-void MainWindow::showMenus(bool show) {
-  menuBar()->setVisible(show);
-  fileToolBar->setVisible(show);
-  debugToolBar->setVisible(show);
+void MainWindow::updateMenusVisibility(const bool welcomePageShown) {
```
KochynVolodymyr, line 533: no need const here

KochynVolodymyr (approved):

PR #782, opened 2022-11-04
[RG-161] Extend Help menu with 3 actions. Implemented scrollable licences viewer

TarasSydor:
Given PR implements RG-161 feature: adds 3 extra actions to Help menu and scrollable viewer for licenses,

Documentation and Relese Notes actions aren't implemented yet.
Licenses opens QPlainTextEdit based read-only viewer, containing application specific licenses. Those are supposed to be stored at share/appName/licenses.
It reads all txt files out of that folder and adds the to the view. File contents are split by 3 empty lines separator.
In case such folder doesn't exist (in FOEDAG), Licenses action does nothing,
>

image
image

License viewer takes 2/3 part of the screen

> 
> - [ ] Engine
> - [+] Documentation
> - [ ] Regression tests


src/MainWindow/main_window.cpp
```diff
@@ -1206,3 +1226,12 @@ void MainWindow::onShowStopMessage(bool showStopCompilationMsg) {
...
   m_settings.setValue(SHOW_STOP_COMPILATION_MESSAGE_KEY, m_askStopCompilation);
 }
+
+void MainWindow::onShowLicenses() {
+  auto licensePath = GlobalSession->Context()->DataPath() / LICENSES_DIR;
+  if (!FileUtils::FileExists(licensePath)) return;
+
+  LicenseViewer viewer(licensePath, m_projectInfo.name, this);
+  centerWidget(viewer);
+  viewer.exec();
```
KochynVolodymyr, line 1236: Qt documents says:
Note: Avoid using this function; instead, use open(). Unlike exec(), open() is asynchronous, and does not spin an additional event loop. This prevents a series of dangerous bugs from happening (e.g. deleting the dialog's parent while the dialog is open via exec()). When using open() you can connect to the finished() signal of QDialog to be notified when the dialog is closed.

KochynVolodymyr: but maybe it is fine in this case

TarasSydor:
> @TarasSydor please provide screenshot with whole dialog but not a part.

that was entire dialog, maybe just without a title :)
added a screenshot of entire application

src/MainWindow/main_window.cpp
TarasSydor: for some reason neither show() nor open() show the dialog, so I prefer keeping it as is.
About dialog is also shown via exec().

skyler-rs (approved):

KochynVolodymyr:
@ravic-rs I think for the Licenses window, button Close or Ok should be available at the bottom since all window like this have some button.

TarasSydor:
@ravic-rs 
It seems the original issue (EDA-487) got updated and there is no 'License' action in Help menu anymore:

- Release notes (I am assuming that this also includes all the known issues and bug fixes)
- User guide
- Videos and Tutorials
- Quick Help guide (if we have it available or we can have Tony’s quickstart guide here)

Do we still need to see licenses in the application?

TarasSydor:
> @ravic-rs I think for the Licenses window, button Close or Ok should be available at the bottom since all window like this have some button. Close window by press Enter will not work without button.

close button added (+ screenshot updated)

src/MainWindow/main_window.cpp
ravic-rs: @KochynVolodymyr are you still waiting for this to be resolved ?

PR #800, opened 2022-11-14
verilator support

alain-rs:


src/Compiler/Compiler.h
```diff
@@ -221,7 +225,9 @@ class Compiler {
   /* Compiler class utilities */
   bool RunBatch();
   bool RunCompileTask(Action action);
-  virtual bool ExecuteSystemCommand(const std::string& command);
+
+  virtual void SetEnvironmentVariable(const std::string variable,
```
KochynVolodymyr, line 229: Are you missed references here?

alain-rs: No. It's on purpose.

PR #806, opened 2022-11-16
[GEMINIEDA-550] Added interfaces/classes for structured reports

ravic-rs:
Current PR contains UI part for structured reports and basic parser for Synthesis log file.

Approach and design overview can be found on corresponding confluence page. The idea is to parse the log file line by line, finding some keywords and then recognizing report data.
For the initial implementation, some things were hardcoded (like skipping 3 lines after Printing Statistics to get directly to stats).
Those shortcuts should be improved further.
>
In terms of UI, after compiling the task user can, next to viewing entire log, open structured report with only relevant data
image

Triggering last action runs file parser retrieving relevant data out of the log file. Synthesis report looks as follows:
image

This PR only covers Synthesis report.


src/Compiler/Reports/ITaskReportManager.h
```diff
@@ -0,0 +1,48 @@
...
+#include <map>
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace FOEDAG {
+
+class ITaskReport;
+
+/* Manager for task reports. It has to be implemented per comlilation
```
skyler-rs, line 32: comlilation -> compilation

ravic-rs:
@TarasSydor can we center-align the "Value" column in table. The first column can be left-aligned.
image

src/Compiler/Reports/SynthesisReportManager.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+  return res;
+}
+
+SynthesisReportManager::LinesData SynthesisReportManager::getLevels(
+    const QString &line) const {
+  auto res = LinesData{};
+
+  auto lineData = line.split(" ");
+  res.push_back({MAX_LVL_STR, lineData[10].toStdString()});
```
skyler-rs, line 88: `lineData[10]` is ok for initial parse implementation until we setup a regex solution down the road, but we should wrap these two lines in a conditional to ensure that the hardcoded indexing we are doing doesn't go out of bounds on `lineData`. Probably not an actual concern since we know what's being output, but doesn't hurt to add some protective code in case the output changes down the road.

src/Compiler/Reports/SynthesisReportManager.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+  // To save the last report statistics
+  auto stats = LinesData{};
+  auto levels = LinesData{};
+
+  QTextStream in(&logFile);
+  QString line;
+  while (in.readLineInto(&line)) {
+    auto dataLine = line.simplified();
+    if (dataLine.contains(QString(STAT_STR))) {
```
skyler-rs, line 114: These two checks for `STAT_STR` and `DE_STR` might be a bit more efficient if you check for (DE_STR) first and then then do the `STAT_STR` check in an `else if` so we aren't double searching each line and we're cancelling out as early as possible if we find `DE_STR` at the beginning of a line.

If we see slow down from the `STAT_STR` check, assuming `STAT_STR` always starts at the same char position on a line, you might be able to substring a hardcoded portion of the line and compare against that... assuming substring is faster than searching the whole line.

ravic-rs:
Some issues to address:
1. macos checks.
2. Refactoring for regex.
3. Using the status bar to let users know that reports are been generated (especially for bigger log files).
4. Testing this on a biggest design possible for performance check.

src/Compiler/TaskTableView.cpp
```diff
@@ -175,12 +175,20 @@ void TaskTableView::addTaskLogAction(QMenu *menu, FOEDAG::Task *task) {
   QString viewLogStr = "View " + title + " Logs";
   QAction *viewLog = new QAction(viewLogStr, this);
   logFilePath.replace(PROJECT_OSRCDIR, Project::Instance()->projectPath());
-  if (!QFile(logFilePath).exists()) {
+  auto logExists = QFile(logFilePath).exists();
```
KochynVolodymyr, line 178: there is static member in QFile for this.

src/Compiler/TaskManager.h
```diff
@@ -86,6 +87,7 @@ class TaskManager : public QObject {
 
   void setTaskCount(int count);
 
+  TaskReportManagerRegistry &getReportManagerRegistry();
```
KochynVolodymyr, line 90: maybe const reference?

src/Compiler/Reports/TaskReportManagerRegistry.h
```diff
@@ -0,0 +1,46 @@
...
+#include <memory>
+#include <unordered_map>
+
+#include "ITaskReportManager.h"
+
+namespace FOEDAG {
+
+class TaskReportManagerRegistry final {
+ public:
+  using ReportManager = std::unique_ptr<ITaskReportManager>;
```
KochynVolodymyr, line 34: could you change name to something like ReportManager_ptr ? would be nice to see that it is actually start pointer

src/Compiler/Reports/ITaskReport.h
```diff
@@ -0,0 +1,43 @@
...
+/* Given interface represents a report for the compilation task.
+ * It returns row and column data as strings. For simplicity, only
+ * standard table data is supported, no parent items. It can be
+ * extended if needed.
+ */
+class ITaskReport {
+ public:
+  using LineValues = std::vector<std::string>;
+  // Returns report column names
+  virtual const LineValues &getColumns() const = 0;
```
KochynVolodymyr, line 37: virtual destructor is missing

src/Compiler/Reports/TableReport.h
```diff
@@ -0,0 +1,46 @@
...
+#pragma once
+
+#include "ITaskReport.h"
+
+namespace FOEDAG {
+
+// Default report implementation, representing table-like data.
+class TableReport : public ITaskReport {
+ public:
+  TableReport(LineValues &&columnNames, std::vector<LineValues> &&linesData,
```
KochynVolodymyr, line 30: Move constructor should be provided if you really care about performance

src/Compiler/Reports/ITaskReportManager.h
```diff
@@ -0,0 +1,48 @@
...
+class ITaskReport;
+
+/* Manager for task reports. It has to be implemented per compilation
+ * task, as reports are task-specific. It knows how many reports are
+ * available per task and can create reports.
+ */
+class ITaskReportManager {
+ public:
+  // Returns all available reports per compilation task
+  virtual std::vector<std::string> getAvailableReportIds() const = 0;
```
KochynVolodymyr, line 39: virtual destructor should be defined since you declare an interface

src/Compiler/Reports/SynthesisReportManager.cpp
TarasSydor: I agree

TarasSydor:
> Some issues to address:
> 
> 1. macos checks.
> 2. Refactoring for regex.
> 3. Using the status bar to let users know that reports are been generated (especially for bigger log files).
> 4. Testing this on a biggest design possible for performance check.

1. Fixed
2. Still to come
3. Fixed
4. Creating a report for 15k lines file takes nothing. I increased the file size to 120k (by simply copy pasting), but still there is no delay, not even fraction of a second.

ravic-rs:
@volodymyrkochyn please mark your comments as resolved, if that's true.
@TarasSydor there's a conflict.

src/Compiler/Reports/SynthesisReportManager.cpp
ravic-rs: @skyler-rs can we mark this as resolved ?

ravic-rs:
TODOs
1. Merge the vectors to bottom and top.

ravic-rs:
> TODOs
> 
> 1. Merge the vectors to bottom and top.

This is done.

PR #815, opened 2022-11-19
[RG-185] Add support for post synthesis project

KochynVolodymyr:
Add post synthesis project type

- [x] Library: compiler, newproject
- [ ] Engine
- [ ] Documentation
- [ ] Regression tests

### Screenshots
image
image
image

### DIFF
--- a/src/Compiler/TaskModel.cpp
+++ b/src/Compiler/TaskModel.cpp
@@ -48,7 +48,16 @@ bool TaskModel::hasChildren(const QModelIndex &parent) const {
 }
 
 Qt::ItemFlags TaskModel::flags(const QModelIndex &index) const {
-  return QAbstractItemModel::flags(index);
+  auto taskId = ToTaskId(index);
+  auto task = m_taskManager ? m_taskManager->task(taskId) : nullptr;
+  auto flags = QAbstractItemModel::flags(index);
+  if (task) {
+    if (task->isEnable())
+      flags |= Qt::ItemIsEnabled;
+    else
+      flags &= ~Qt::ItemIsEnabled;
+  }
+  return flags;
 }
 
 uint TaskModel::ToTaskId(const QModelIndex &index) const {
--- a/src/Main/ProjectFile/ProjectManagerComponent.cpp
+++ b/src/Main/ProjectFile/ProjectManagerComponent.cpp
@@ -71,7 +71,7 @@ void ProjectManagerComponent::Save(QXmlStreamWriter* writer) {
 
   stream.writeStartElement(PROJECT_OPTION);
   stream.writeAttribute(PROJECT_NAME, PROJECT_CONFIG_TYPE);
-  stream.writeAttribute(PROJECT_VAL, tmpProCfg->projectType());
+  stream.writeAttribute(PROJECT_VAL, QString::number(tmpProCfg->projectType()));
   stream.writeEndElement();
 
   QMap<QString, QString> tmpOption = tmpProCfg->getMapOption();
@@ -223,8 +223,9 @@ void ProjectManagerComponent::Load(QXmlStreamReader* r) {
                   reader.attributes().value(PROJECT_VAL).toString());
             } else if (PROJECT_CONFIG_TYPE ==
                        reader.attributes().value(PROJECT_NAME).toString()) {
-              tmpProCfg->setProjectType(
-                  reader.attributes().value(PROJECT_VAL).toString());
+              bool ok{true};
+              auto type = reader.attributes().value(PROJECT_VAL).toInt(&ok);
+              tmpProCfg->setProjectType(ok ? type : RTL);
             } else {
               tmpProCfg->setOption(
                   reader.attributes().value(PROJECT_NAME).toString(),
--- a/src/NewProject/ProjectManager/project_configuration.cpp
+++ b/src/NewProject/ProjectManager/project_configuration.cpp
@@ -7,7 +7,7 @@ using namespace FOEDAG;
 ProjectConfiguration::ProjectConfiguration(QObject *parent)
     : ProjectOption(parent) {
   initProjectID();
-  m_projectType = "";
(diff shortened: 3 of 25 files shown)


src/Compiler/TaskModel.cpp
```diff
@@ -48,7 +48,16 @@ bool TaskModel::hasChildren(const QModelIndex &parent) const {
...
 Qt::ItemFlags TaskModel::flags(const QModelIndex &index) const {
-  return QAbstractItemModel::flags(index);
+  auto taskId = ToTaskId(index);
+  auto task = m_taskManager ? m_taskManager->task(taskId) : nullptr;
+  auto flags = QAbstractItemModel::flags(index);
+  if (task) {
+    if (task->isEnable())
+      flags |= Qt::ItemIsEnabled;
+    else
+      flags &= ~Qt::ItemIsEnabled;
```
TarasSydor, line 58: Does it also disable context menu for the task?

src/NewProject/ProjectManager/project_configuration.cpp
```diff
@@ -7,17 +7,17 @@ using namespace FOEDAG;
 ProjectConfiguration::ProjectConfiguration(QObject *parent)
     : ProjectOption(parent) {
   initProjectID();
-  m_projectType = "";
+  m_projectType = 0;  // RTL
```
TarasSydor, line 10: can we make project type member an enum?

src/Main/ProjectFile/ProjectManagerComponent.cpp
```diff
@@ -71,7 +71,7 @@ void ProjectManagerComponent::Save(QXmlStreamWriter* writer) {
 
   stream.writeStartElement(PROJECT_OPTION);
   stream.writeAttribute(PROJECT_NAME, PROJECT_CONFIG_TYPE);
-  stream.writeAttribute(PROJECT_VAL, tmpProCfg->projectType());
+  stream.writeAttribute(PROJECT_VAL, QString::number(tmpProCfg->projectType()));
```
TarasSydor, line 74: how about serializing project type by name, but not its number? 
we could make use of helper function converting project type to string (currently part of project_type_form).

src/NewProject/project_type_form.cpp
```diff
@@ -14,27 +14,46 @@ projectTypeForm::projectTypeForm(QWidget *parent)
...
+    if (ui->radioBtnPOSMixed->isChecked()) {
+      return PostSynthWithHDL;
+    } else if (ui->radioBtnPOSPure->isChecked()) {
+      return PostSynthPure;
+    }
+  }
+  return RTL;
+}
+
+QString projectTypeForm::projectTypeStr() const {
```
TarasSydor, line 49: we may need this logic elsewhere. helper function can be moved to some generic place. we can also add one more for backwards conversion. 
in this case we can save project name into the project file, which is safer than just an integer.

TarasSydor (changes requested):

src/NewProject/ProjectManager/project_configuration.cpp
KochynVolodymyr: We have it, but not here. I don't want to spread dependencies.

src/Main/ProjectFile/ProjectManagerComponent.cpp
KochynVolodymyr: it is always better to use numbers and not strings

TarasSydor (approved):

ravic-rs:
@KochynVolodymyr got some conflicts to resolve.

KochynVolodymyr:
@ravic-rs resolved

src/NewProject/project_type_form.cpp
```diff
@@ -14,27 +14,50 @@ projectTypeForm::projectTypeForm(QWidget *parent)
...
+  return projectTypeStr(projectType());
+}
+
+QString projectTypeForm::projectTypeStr(ProjectType type) {
+  switch (type) {
+    case RTL:
+      return "RTL";
+    case PostSynthWithHDL:
+      return "Post-synthesis with HDL";
```
ravic-rs, line 58: We should show this in the "Summary" to distinguish the two sub-types.

ravic-rs:
For the POS 2.2 where user can only add one edif or eblif or blif. 

1. The "Create File" option needs an update.
From one of the previous cases we removed the templates, should we just create "empty" supported files in this case ?
image

2. We also need to update the language selection to include only the supported types.
image

src/NewProject/project_type_form.cpp
KochynVolodymyr: We do.
image

KochynVolodymyr:
> The "Create File" option needs an update.

I didn't get it.

> We also need to update the language selection to include only the supported types.

It make sense. I'll push update

KochynVolodymyr:
@alain-rs  could you please clarify this?
image

KochynVolodymyr:
> For the POS 2.2 where user can only add one edif or eblif or blif.

@ravic-rs Resolved

alaindargelas:
Netlist file types are: BLIF, EBLIF, EDIF, Verilog

alaindargelas:
The Post Synthesis project does not need 2 sub options. 
There is only one choice: add one netlist file, please remove the sub options.

ravic-rs:
@volodymyrkochyn can we update the screenshot in description?
And text option as
-> User can only add one edif or .ebif or .blif file
     - Run P&R timing & generate bitstream

alaindargelas:
Change: User can only add one edif or .ebif or .blif file
To:  Only one of .edif, .blif, .eblif, .v file allowed.

alaindargelas:
Yes that one too, it is the same as .edif

KochynVolodymyr:
@alaindargelas please confirm that compilation process starts with Place and Route. Synth disabled.

alaindargelas:
Correct

PR #817, opened 2022-11-21
RG-180: adding wave_* cmds to launch and control GTKWave

skyler-rs:
Note: CI appears to be failing with a Qt5 package issue currently. It is highly unlikely this PR is causing that.

> - [x] To address an existing issue. If so, please provide a link to the issue: RG-180

> - gtkwave was launched from the console with `gtkwave` which could cause some confusion about if a local or system copy of gtkwave was being launched
>
> - gtkwave process is now launched with --wish which allows for remote control via tcl
> - gtkwave process is launched (once) from any of the wave_* commands below
>   - There is logic to track the process and send all commands to the current gtkwave process. If the gtkwave window is closed, a new process is started on next command and all subsequent commands will communicate with the new process
> - foedag tcl commands are provided to help control the gtkwave process:
>   - Note: all commands operate with the existing gtkwave process or open a new one if none is present
>   - `wave_open <fileName>` load the given file in gtkwave
>   - `wave_cmd <cmdStr>` send tcl commands to the gtkwave process see [gtkwave pdf for gtkwave:: commands](https://gtkwave.sourceforge.net/gtkwave.pdf)
>   - `wave_show <signalName>` adds a signal if it hasn't been added already and highlights the signal, any previous highlights are cleared
>   - `wave_time <timeStr>` sets the left side of the gtkwave window to the time specified. Note: time units can be specified (w/o a space) ex: 1ns, 1000ps, etc
>   - `wave_refresh` this will reload the waveform file, this is the equivalent of calling `wave_cmd gtkwave::reloadFile`
>   - `wave_get_return` this will retrieve the last return value stored by the gtkwave process handler. A return value is stored if it is prefixed with `_RETURN_`. See gtkwave_invoke.tcl test for usage.
> - Additional helper commands are installed into the GTKWave tcl interpreter to simplify some commands:
>   - `foedag::unselect_all` unselects all signals w/o having to provide the names of the signals to de-select
>   - `foedag::add_signal_once` adds a signal, but only if it hasn't been added already
>   - `foedag::show_signal` adds a signal once, clears previous highlights, and selects the given signal

> 
> - [x] Library: Compiler
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

> ### Testing
- Manually tested in foedag and raptor
- added tests/TestGui/gtkwave_invoke.tcl to ensure that gtkwave is invoked and can be communicated with (Note, due to noisy CI runs, this test will not be added to the makefile for now)


src/Compiler/Compiler.h
```diff
@@ -203,6 +205,10 @@ class Compiler {
   void SetWaveformFile(const std::string& wave) { m_waveformFile = wave; }
   const std::string& GetWavefromFile() { return m_waveformFile; }
 
+  QProcess* GetGTKWaveProcess();
+  void GTKWave_send_cmd(const std::string& gtkWaveCmd,
```
KochynVolodymyr, line 209: Using underscore _ is out of our coding convention.

src/Compiler/Compiler.cpp
```diff
@@ -1461,33 +1461,238 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
...
+    m_gtkwave_process->start("./gtkwave", args);
+    if (m_gtkwave_process->waitForStarted()) {
+      // Clear pointer on process close
+      QObject::connect(
+          m_gtkwave_process,
+          // static_cast required due to overload of signal params
+          static_cast<void (QProcess::*)(int, QProcess::ExitStatus)>(
+              &QProcess::finished),
+          [this](int exitCode, QProcess::ExitStatus exitStauts) {
+            m_gtkwave_process = nullptr;
```
KochynVolodymyr, line 1617: memory leak is here

src/Compiler/Compiler.cpp
```diff
@@ -1461,33 +1461,238 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
...
+      if (argc != 2) {
+        Tcl_AppendResult(
+            interp,
+            qPrintable("Expected Syntax: wave_time <timeString>\ntimeString "
+                       "can specify its time units. Ex: 1ps, 1000ns, 1s"),
+            nullptr);
+        return TCL_ERROR;
+      }
+
+      QString cmd =
```
KochynVolodymyr, line 1523: we try do not use QString for the compiler if it is possible. In case tcl ,commands I guess we can use std::string

src/Compiler/Compiler.cpp
```diff
@@ -1461,33 +1461,238 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
...
+    Compiler* compiler = (Compiler*)clientData;
+    if (compiler) {
+      QString file{};
+      // check if a file was requested
+      if (argc > 1) {
+        file = QString::fromStdString(argv[1]);
+        if (file.count() > 0 && file[0] == "~") {
+          // ~/ doesn't get substitued so we'll manually turn ~ paths into
+          // absolute paths
+          file = QDir::homePath() + file.mid(1);
```
KochynVolodymyr, line 1493: didn't you forget slash / here?

skyler-rs: The scenario was ~/ so file.mid(1) would return the / as well, but i did realize there is a rare scenario where a user could provide a path like ~userName/file.txt which wouldn't work in this case so i've updated the logic.

src/Compiler/Compiler.cpp
skyler-rs: fixed

src/Compiler/Compiler.cpp
```diff
@@ -1461,33 +1461,248 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
...
+                     const char* argv[]) -> int {
+    Compiler* compiler = (Compiler*)clientData;
+    if (compiler) {
+      // Convert args to a single string
+      QStringList args{};
+      for (int i = 1; i < argc; i++) {
+        args.append(argv[i]);
       }
-      args << file;
+      QString cmd = args.join(" ") + "\n";
```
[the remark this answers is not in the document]
skyler-rs: updated to std::string

src/Compiler/Compiler.cpp
```diff
@@ -1461,33 +1461,248 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
...
+
+    // Get a path to our local gtkwave binary
+    auto binPath = GlobalSession->Context()->BinaryPath();
+    auto exePath = binPath / "gtkwave" / "bin";
+    QString wd = QString::fromStdString(exePath.string());
+    // Need to be in the gtkwave bin dir to resolve some dependency issues
+    m_gtkwave_process->setWorkingDirectory(wd);
+
+    // Enable process control via tcl interface over std in with --wish
+    QString wishArg = "--wish";
```
ravic-rs, line 1610: Same stdd:string question here.

src/Compiler/Compiler.cpp
```diff
@@ -1461,33 +1461,248 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
...
+            // getters, this is where we should check the value.
+
+            // Possible way of returning getter results:
+            // `wave_cmd puts \[gtkwave::getWaveWidth\]`
+            // also see installGTKWaveHelpers() for fine control
+
+            // Read stdout data
+            QByteArray data = m_gtkwave_process->readAllStandardOutput();
+            QString trimmed = data.trimmed();
+            // std::cout << "stdout: " << trimmed.toStdString() << std::endl;
```
ravic-rs, line 1643: QString question, plus left over debug statement ?

src/Compiler/Compiler.cpp
```diff
@@ -1461,33 +1461,248 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
...
+            if (trimmed.startsWith("Interpreter id is")) {
+              // Install extra tcl helpers
+              installGTKWaveHelpers();
+            }
+          });
+
+      // Listen to stderr
+      QObject::connect(m_gtkwave_process, &QProcess::readyReadStandardError,
+                       []() {
+                         // TODO @skyler-rs nov2022 This should be logged once
```
ravic-rs, line 1655: We do have a logging module, which one are we waiting on ?

src/Compiler/Compiler.cpp
```diff
@@ -1461,33 +1461,248 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
...
+                         // we have a logging system
+
+                         // QByteArray data =
```
ravic-rs, line 1658: Debug ?

src/Compiler/Compiler.cpp
skyler-rs: QProcess is Qt and it uses a QStringList which wishArg gets stored in so we could use std::string for a line, but we are going to have to use QString eventually...

src/Compiler/Compiler.cpp
skyler-rs: We are reading directly from a Qt interface, there is no benefit from converting it to std::string

src/Compiler/Compiler.cpp
skyler-rs: Yes, we don't have a way of logging errors just for devs so this gets uncommented often when debugging.

src/Compiler/Compiler.cpp
skyler-rs: I mean a logging system where we can capture all the extra debug info that will help devs but is overkill for the user. GTKWave uses stderr more than just errors so it's not worth it to report errors to the user currently.

ravic-rs:
Was this tested by running a tcl script or invoking using the tcl console ?

skyler-rs:
Tcl console

ravic-rs (approved):
Looks good at the console. We need a way to verify the new tcl cmd set work as expected. Although verifying an external application behavior is out of scope.

skyler-rs:
> Looks good at the console. We need a way to verify the new tcl cmd set work as expected. Although verifying an external application behavior is out of scope.

This is possible, but it will require some additional code to come up w/ a way to get return values back from gtkwave, working on it now.

skyler-rs:
Added gtkwave_invoke.tcl which will ensure the gtkwave process is launched and can be communicated with.

PR #818, opened 2022-11-22
Fixed wrong file type when file extension is in upper case

KochynVolodymyr:
When user add files we detect wrong file type in case file extension is in upper case.

- [x] Library: compiler
- [ ] Engine
- [ ] Documentation
- [x] Regression tests

### Screenshots
image

### DIFF
--- a/src/Compiler/CompilerDefines.cpp
+++ b/src/Compiler/CompilerDefines.cpp
@@ -31,6 +31,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "TaskManager.h"
 #include "TaskModel.h"
 #include "TaskTableView.h"
+#include "Utils/QtUtils.h"
 
 extern FOEDAG::Session *GlobalSession;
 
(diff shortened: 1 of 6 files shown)


src/Compiler/CompilerDefines.cpp
```diff
@@ -140,3 +142,7 @@ bool FOEDAG::target_device(const QString &target) {
                            qPrintable(QString("target_device %1").arg(target)));
   return (res == TCL_OK);
 }
+
+bool FOEDAG::IsEqual(const QString &str, const QString &s) {
```
ravic-rs, line 146: This sounds like a place for the utils class, no ?

KochynVolodymyr: Done

ravic-rs: Why not the string utils ? How are we planning to differentiate between QtUtils and StringUtils ?
https://github.com/os-fpga/FOEDAG/blob/main/src/Utils/StringUtils.cpp

KochynVolodymyr: StringUtils related to std::string. If we have something that related to Qt it should go to QtUtils

PR #820, opened 2022-11-22
Update IP Generator to "quote" string types

skyler-rs:
> - [x] Latest IP_Catalog changes expect string type for some fields which requires an update in the foedag ip generator output

> - IP Generator prints values unquoted
>
> - IP Generator has been updated to look up lost parameter info from the ip catalog so types like strings and filetypes can be quoted liked strings in the json output

> 
> - [x] Library: IPGenerate
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

> ### Testing
> - Manually tested in foedag and raptor. In Raptor, tested all IPs to ensure generate actually creates files etc.
> - *_gen.py file updates are small changes to add string types for manual testing


src/IPGenerate/IPGenerator.cpp
```diff
@@ -427,18 +450,30 @@ bool IPGenerator::Generate() {
...
-              break;
-            case Value::Type::ParamString:
-              value = param.GetSValue();
-              break;
-            case Value::Type::ParamInt:
-              value = param.GetSValue();
-              break;
-            case Value::Type::ConstInt:
-              value = param.GetSValue();
+          // configure ip loses type info so we'll use the ip catalog's
```
ravic-rs, line 453: Can you we elaborate it with the reasons for the loss of type?

ravic-rs (commented):
1 comment to address.

src/IPGenerate/IPGenerator.cpp
skyler-rs: updated

ravic-rs (approved):

PR #821, opened 2022-11-22
RG-191: Implemented report manager for Placement

TarasSydor:
Given PR fulfills RG-191 and introduces structured reports for Placement task.

Parses goes through the log file, locating familiar lines and creates two reports:
- Timing, which is placed into a separate file (post_place_timing.rpt)
- Resource utilization, shown in application editor.
Due to cumbersome data structure in a log file we couldn't locate the chunk of data to parse it separately. Instead we go through all the lines, comparing it with key regexps. Once familiar one is met, we either retrieve entire line for timing report or start resources parser.
As timing reporting will be changes after OpenSTA migration, we just place not formatted lines into the file - the logic will have to be revised anyway.
>
Timing report file generated:
image

Resource utilization report opened:
image

> 
> - [ ] Engine
> - [+] Documentation
> - [ ] Regression tests


src/Compiler/Reports/PlacementReportManager.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+namespace {
+static constexpr const char *REPORT_NAME{"Report Resource Utilization"};
+static constexpr const char *RESOURCES_SPLIT{"blocks of type:"};
+
+static constexpr const char *BLOCKS_COL{"Blocks"};
+static constexpr const char *NETLIST_COL{"Netlist"};
+static constexpr const char *ARCHITECTURE_COL{"Architecture"};
+
+static const QRegExp FIND_TIMINGS{
```
skyler-rs, line 40: When we revisit the reports, we should consider switching to QRegularExpression as QRegExp class is deprecated in Qt6. Additionally, QRegularExpression has better support for perl regex so the syntax issues you were bumping into might be resolved by using QRegularExpression instead.

ravic-rs: Why can't we use now? Wondering if it's a lot of refactoring. We got other reports to build on.

src/Compiler/Reports/PlacementReportManager.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+std::unique_ptr<ITaskReport> PlacementReportManager::createReport(
+    const std::string &reportId) {
+  auto projectPath = Project::Instance()->projectPath().toStdString();
+  auto logPath =
+      std::filesystem::path(projectPath) / std::string(PLACEMENT_LOG);
+
+  if (!FileUtils::FileExists(logPath)) return nullptr;
+
+  auto logFile = QFile(QString::fromStdString(logPath.string()));
```
KochynVolodymyr, line 59: can we avoid mixing QString and std::string? I see here a lot of this

src/Compiler/Reports/PlacementReportManager.cpp
skyler-rs: It sounded like a number of these parsers are going to get some level of rewrite once the reporting format is more determined so i figured it'd be easier to make the changes then instead of introducing new logic mid PR as that would require re-testing etc.
Not sure how much of a change it is, [this section ](https://doc.qt.io/qt-5/qregularexpression.html#notes-for-qregexp-users) makes it look like there are a couple changes to consider.

src/Compiler/Reports/PlacementReportManager.cpp
ravic-rs: +1

src/Compiler/Reports/PlacementReportManager.cpp
ravic-rs: Thanks.

ravic-rs (changes requested):
Please resolve the CI failures.
https://github.com/os-fpga/FOEDAG/actions/runs/3527055140/jobs/5915686178#step:9:1165

src/Compiler/Reports/PlacementReportManager.cpp
TarasSydor: yeah. I used std::string in ITaskReport class, but files are parsed via Qt. it caused lots of conversions back and forth.
now the interface works on Qt classes and conversions were needless. I also removed them from synthesis parser.

PR #825, opened 2022-11-24
Implemented report manager for Routing task.

TarasSydor:
This PR implements RG-`192 task and introduced structured reports for Routing compilation task.

Routing actually has two reports: Resource Utilization, which is similar to Placement one, and Circuit statistics. In order not to duplicate the parser code an AbstractReportManager was introduced, which has some common logic for parsing.
Circuit statistics only covers part of entire table (Blocks section).

Two reports are opened in separate tabs. As context menu is a temporal solution and we'll get Reports tab anyway, it didn't make sense to increase the view complexity and add multi-view support. User will only be able to open a single report per tab.
>
image

image

> 
> - [ ] Engine
> - [+] Documentation
> - [ ] Regression tests


KochynVolodymyr (approved):

skyler-rs (approved):

ravic-rs (approved):
Looks good.

PR #827, opened 2022-11-25
[RG-159] Allow refresh pin planner tables after *.pin file was changed on the disk

KochynVolodymyr:
Allow user refresh pin planner in case *.pin constraint file was changed on disk or by user or some other app.

- [x] Library: pinassignment, foedagcore, texteditor
- [ ] Engine
- [ ] Documentation
- [x] Regression tests

### Screenshots
image
image

### DIFF
--- a/src/MainWindow/main_window.cpp
+++ b/src/MainWindow/main_window.cpp
@@ -298,32 +298,33 @@ void MainWindow::startStopButtonsState() {
 
 QDockWidget* MainWindow::PrepareTab(const QString& name, const QString& objName,
                                     QWidget* widget, QDockWidget* tabToAdd,
-                                    Qt::DockWidgetArea area,
-                                    bool refreshButton) {
+                                    Qt::DockWidgetArea area) {
   QDockWidget* dock = new QDockWidget(name, this);
   dock->setObjectName(objName);
   dock->setWidget(widget);
-  if (refreshButton) {
-    auto btn = new QPushButton;
-    connect(btn, &QPushButton::clicked, this, &MainWindow::refreshPinPlanner);
-    btn->setSizePolicy(QSizePolicy{QSizePolicy::Maximum, QSizePolicy::Maximum});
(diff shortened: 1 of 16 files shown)


src/MainWindow/main_window.cpp
```diff
@@ -1220,6 +1271,24 @@ bool MainWindow::confirmExitProgram() {
               QMessageBox::No | QMessageBox::Yes) == QMessageBox::Yes);
 }
 
+void MainWindow::setVisibleRefreshButtons(bool visible) {
```
skyler-rs, line 1274: Should probably rename to indicate this only works with pin planner docks. Anything that uses PrepareTab can add this refresh button so currently the function name seems to imply it will update all refresh buttons.

src/MainWindow/main_window.cpp
```diff
@@ -1220,6 +1271,24 @@ bool MainWindow::confirmExitProgram() {
...
+  for (auto dock : m_pinAssignmentDocks) {
+    auto button = dock->findChild<QPushButton*>();
```
skyler-rs, line 1276: dock->findChild<QPushButton*> will return whatever first button it finds which is probably ok in a limited scenario where we know the widget construction, but this could have unexpected results down the road. Might want to set the objectName of your refresh button then update findChild to `findChild<QPushButton*>("refreshBtnName")`

TarasSydor (approved):

src/MainWindow/main_window.cpp
KochynVolodymyr: PrepareTab updated. Now it doesn't related to pin planner.

src/MainWindow/main_window.cpp
KochynVolodymyr: It makes sense, updated!

skyler-rs (approved):

ravic-rs (approved):
Looks good.

PR #834, opened 2022-11-29
[RG-204] Introduced Reports tab, appearing next to the Console after running a project

TarasSydor:
Given PR adds extra Reports tab, requested by RG-204.

The tab UI is implemented as ReportsTreeWidget and is based on QTreeWidget. QTreeView was also an option, but it requires more coding and as amount of tasks or reports will not gonna change frequently, QTreeWidget is sufficient.

Next to that, ITaskReportManager API was moved from std::string to QString so as to avoid extra conversions.
>
After compilation is finished a Reports tab appears in bottom area:
image

The report gets opened after double-clicking on a leaf node. If the report is opened already, it will get selected.

One more change - actions in tasks view are only visible when the task has underlying reports. Each action corresponds to the report - functionally agreed as we don't expect more than 2 reports per task:
image

> 
> - [ ] Engine
> - [+] Documentation
> - [ ] Regression tests


ravic-rs (changes requested):
1, Typo in "Cirtuit"
2. Look at avoiding or replacing the "Reports" in title of tab, widget, column.
3. Some conflicts from other PR merges.

KochynVolodymyr (approved):

skyler-rs (approved):

ravic-rs (approved):
Looks good. We need to handle some usecases such as an empty report/table.

PR #846, opened 2022-12-01
GTKWave: Fix interface issues and change non-error msg to status msg.

skyler-rs:
> Fix a typo in the wave_refresh command and convert a normal message that gtkwave sends as an error to a status message instead.

> - `wave_refresh` has a typo and calls `gtkwave::reloadFile` when it should be `gtkwave::reLoadFile`
> - `wave_open` sent relative paths to gtkwave which launches from a different bin dir which could result in files not opening.
> - GTKWave reports a file's start and end time at load, but it does it on stderr for some reason so every load prints an error message:
>   image

>
> - `wave_refresh` now calls `gtkwave::reLoadFile`
> - `wave_open` now converts relative paths to full paths before passing to GTKWave. If the file doesn't exist, foedag will print an error.
> - Load messages are now treated as status instead of error:
>   image

> 
> - [x] Library: Compiler
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

> ### Testing
> - Manually tested in foedag
> - tests/gtkwave_open_bad_path.tcl has been added to verify `wave_open` path fix
> - wave_refresh change will be tested when RG-199 is finished


src/Compiler/Compiler.cpp
```diff
@@ -1539,10 +1539,20 @@ bool Compiler::RegisterCommands(TclInterpreter* interp, bool batchMode) {
         }
       }
 
+      // GTKWave sets its current directory to its bin dir for dependency
```
ravic-rs, line 1542: Just in case this was not considered: GTKWave location will be different on user install.

ravic-rs (approved):
Looks good.

PR #847, opened 2022-12-01
[RG-199] Add test for wave_* tcl commands

skyler-rs:
### Note: This is currently failing due to libgtk-3.so.0 being missing on one of the CI targets, checking with Nadeem on how to resolve.

> - [x] To address an existing issue. If so, please provide a link to the issue: [RG-199]

> - Testing was done manually due to CI issues
>
> - Now that GTKWave stdout and stderr are reported in the console, we can use foedag.log for normal tcl verification
>    - A test has been added to test the various wave_* commands

> 
> - [ ] Engine
> - [ ] Documentation
> - [x] Regression tests

> ### Testing
> - Note: Delays are added after file load and refresh operations because the test.vcd is from a real design that takes a moment to load which could cause the testcase to fail. This is only a failure situation because the test is running so fast that after we request that GTKWave opens a file (asynchronously), we are already grepping the log for the load result before GTKWave has finished loading.


alaindargelas:
@skyler-rs please add a test for wave_open with an actual simulation file:
wave_open <filename> 
From what I tried, that does not work.

skyler-rs:
> @skyler-rs please add a test for wave_open with an actual simulation file:
> wave_open
> From what I tried, that does not work.

Hey @alaindargelas, there is a line that tests that https://github.com/os-fpga/FOEDAG/pull/847/files#diff-ecc05c174feb34b32733c85157a214ef61195444c26ec6d0fbfa84c98a645074R84

test.vcd is a copy of a tb.vcd from the Raptor test dirs.

Note there was an issue w/ wave_open when using relative paths (absolute should work fine if you need to use it now) which will be fixed with https://github.com/os-fpga/FOEDAG/pull/846.

skyler-rs:
This will need some more time to test after the install/uninstall changes are stable. This probably will have some delay issues with our CI that will need debugging. Closing for now to avoid confusion.

PR #852, opened 2022-12-02
[RG-193] Extended main window with Messages tab, holding important compilation messages

TarasSydor:
This PR addresses RG-193 and extends main window with Messages tab.

After compilation is finished, Messages tab gets added next to the console. It holds messages taken from log files. Messages are visualized in a tree view, where parent items represent tasks and leafs - messages.
In order to fill Messages tab, file parsing is triggered.

Double clicking on a message opens corresponding log file and selects the line, from which the message was taken.
>
 Messages tab looks as follows (only Synthesis messages are supported. other ones still to come) 
image

Double clicking on a message opens (or activates) the log file and highlights the corresponding line. View gets scrolled so that line is in the middle of the screen:

image

On top of that, I also included RG-217 into this PR and it's still open. It allows to configure the visibility of Messages and Reports tab in View menu:
image

Tabs are shown for the first after projects compilation, afterwards - according to the View menu. At startup tabs are hidden.


src/TextEditor/editor.cpp
```diff
@@ -1,5 +1,7 @@
 #include "editor.h"
 
+#include <QScrollBar>
```
skyler-rs, line 3: Accidental include? Don't think these changes require this.

src/Compiler/Reports/SynthesisReportManager.h
```diff
@@ -38,14 +40,27 @@ namespace FOEDAG {
...
   // Retrieves maximum and average levels out of given line and fills into stats
   void fillLevels(const QString &line, ITaskReport::TableData &stats) const;
   // Parses input stream and gets all statistics with their values
   ITaskReport::TableData getStatistics(const QString &statsStr) const;
+  // Go through the log file and fills internal data collections (stats,
+  // messages)
+  void parseLogFile();
+
+ private:
+  bool m_fileParsed{false};
```
skyler-rs, line 60: Mac target gives a [pedantic error ](https://github.com/os-fpga/FOEDAG/actions/runs/3604209452/jobs/6073314906#step:10:1236)for this. Since m_fileParsed is in the AbstractReportManager base class, probably just need to remove this line.

skyler-rs (changes requested):
Looks good after Mac CI fix and removing possibly unnecessary include if it isn't needed.

KochynVolodymyr (approved):
Good job!

src/Compiler/Reports/SynthesisReportManager.cpp
```diff
@@ -29,13 +29,22 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "TableReport.h"
 
 namespace {
+// Report strings
 static constexpr const char *REPORT_NAME{"Synthesis report"};
 static constexpr const char *MAX_LVL_STR{"Maximum logic level"};
 static constexpr const char *AVG_LVL_STR{"Average logic level"};
+
+// Messages regexp
+static const QRegExp MESSAGES_REGEXP{
```
ravic-rs, line 38: Just a thought, we can let user comments configure the messages to be extracted via reading a file or settings dialog.

src/MainWindow/MessagesTabWidget.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+  auto &reports = m_taskManager.getReportManagerRegistry();
+  for (auto task : m_taskManager.tasks()) {
+    auto taskId = m_taskManager.taskId(task);
+
+    if (auto reportManager = reports.getReportManager(taskId)) {
+      auto taskItem = new QTreeWidgetItem({task->title()});
+      taskItem->setData(0, Qt::UserRole, task->logFileReadPath());
+
+      const auto &msgs = reportManager->getMessages();
```
ravic-rs, line 32: Did you think about adding user-defined log levels to the messages ? Also, I think it's a good idea to let user know the current log file location in this widget.

ravic-rs (commented):
Couple of comments to address.

src/Compiler/Reports/SynthesisReportManager.cpp
TarasSydor: that might be a powerful features, yes. but sounds rather like a new feature
we may also want to support RegEx there and validate users input.

ravic-rs: Yes it is but keep it in mind since it adds flexibility for each type of user.

src/TextEditor/editor.cpp
TarasSydor: I was experimenting with scrolling and forgot to remove it. fixed

src/Compiler/Reports/SynthesisReportManager.h
TarasSydor: it was added to AbstractReportManager indeed. removed from here

src/MainWindow/MessagesTabWidget.cpp
TarasSydor: I was thinking about various messages "categories" - error, warning info. Will add support for that in one of following PRs.
We can consider User-configurable stuff separately. It shouldn't result in API change, just some extensions here and there.

As for the log file location, I added it next to the task name in parenthesis (see updated PR description screenshot)

ravic-rs:
> Looks good after Mac CI fix and removing possibly unnecessary include if it isn't needed.

@skyler-rs please approve if the CI fixes are ok.

skyler-rs (approved):

skyler-rs:
> > Looks good after Mac CI fix and removing possibly unnecessary include if it isn't needed.
> 
> @skyler-rs please approve if the CI fixes are ok.

Looks like CI is passing now

PR #853, opened 2022-12-02
[RG-218] Add GTKWave Install target

skyler-rs:
> - GTKWave's cmakelists.txt didn't have logic for `make install`
>
> - GTKWave now has an install target
>    - It was observed that after this change, `make uninstall` would not remove gtkwave install as it was being installed to /usr/local/bin/gtkwave and the uninstall scripts only remove foedag and it's share dir.
>       - To ensure gtkwave gets uninstalled as well, gtkwave now installs to the share folder ex: build/share/foedag/gtkwave

> 
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

> ### Testing
> - Changes manually tested in raptor


alaindargelas:
Why are we doing this? All other executable go in the bin directory.

skyler-rs:
> Why are we doing this? All other executable go in the bin directory.

See "What does this pull request change?" in description. I figured moving into share would make the uninstall guaranteed instead of adding explicit uninstall instructions to each repo as our make uninstall seems to only delete foedag folders in /usr/local/*.

alaindargelas:
Incorrect approach. Executable got in bin. Please update the uninstall procedure in each repo.

PR #855, opened 2022-12-06
[RG-220] Fixed few bugs in pin planner

KochynVolodymyr:
Fixed few bugs in pin planner:
* Mode selection cleared when unselect port or pin
* In case unselect all ports or pins mode selection still generate in sdc.
* If unselect pins in IO ports package pin table remains not updated.

 - [x] Library: pinassignment
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests


ravic-rs (approved):
Looks good.

PR #859, opened 2022-12-07
[RG-194, RG-195] Fill Placement/Routing messages.

TarasSydor:
Given PR implements RG-194 and RG-195 features - Placement and Routing messages shown in Messages tab.

Parsing is done line by line. Some logic is placed to Abstract level so that it can be reused by other parsers. Common code parses certain section from a log file, highlighting also some key messages inside. The only requirement is - end of section line starts with the same string as its beginning.

One more important change - TaskMessage structure, which allows multi-level trees and setting a message severity.
image

Some aspects of behavior:
- 4 categories added - ERROR, WARNING, INFO, NONE (used for sections and task items)
- In case no warnings or errors found, no message is shown
- In case multiple warnings/errors were found, double clicking on a corresponding message navigates to the first occurred message/error:
image


src/Compiler/Reports/ITaskReportManager.h
```diff
@@ -26,6 +26,20 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 namespace FOEDAG {
 
+struct TaskMessage {
+  enum class MessageSeverity {
```
KochynVolodymyr, line 30: I think enum class could be moved outside the struct to avoid long access string.

KochynVolodymyr (approved):

skyler-rs (approved):

src/Compiler/Reports/ITaskReportManager.h
TarasSydor: agree. fixed

TarasSydor:
After team validation of Messages tab we came up with some remarks to existing behavior, but it will come in a separate PR.
- Timing report should be created for Placement;
- Messages tab should contain all errors and messages (hidden under collapsible parent item);
- If no errors/warnings found, ''0 errors found" item should not be added;
- Warnings should be grouped. Info messages get sorted by line number;
- Nice to have: collapse/expand all, configuration to hide/show warnings/errors.

TarasSydor:
Changes added by latest commit:
- Messages tab now contains all errors/warnings. Actual messages are shown without an icon within collapsible parent item;
- Messages are sorted by the line number. Warning/Error parent items line number is determined by the first warning/error in a group;
- Synthesis errors/warnings are reported and grouped in a similar to others way: parent item, encapsulating warnings/errors.
- Fixed the issue with Placement resource usage report.

src/Compiler/Reports/AbstractReportManager.cpp
```diff
@@ -106,16 +104,18 @@ int AbstractReportManager::parseErrorWarningSection(QTextStream &in, int lineNr,
 
   QString line;
   // Store line numbers in set to always know which one was first
-  std::set<int> wrnLines, errLines;
+  std::map<int, QString> warnings, errors;
```
KochynVolodymyr, line 107: Are you sure you didn't reorder the elements by using map/set? As we discussed we should keel the order of elements.

TarasSydor: Yes. both std::set and std::map use std::less<Key> comparator by default.
So we end up with the earliest message/error going first in the container.

TarasSydor: Recent commit implements following grouping:
image

Errs/warnings, that aren't split by any Info message, are grouped within a single item. For standalone warnings/errors no grouping is made.

src/Compiler/Reports/AbstractReportManager.cpp
```diff
@@ -128,28 +128,43 @@ int AbstractReportManager::parseErrorWarningSection(QTextStream &in, int lineNr,
...
 
+TaskMessage AbstractReportManager::createWarningErrorItem(
+    MessageSeverity severity, const MessagesLines &msgs) const {
+  auto itemName = severity == MessageSeverity::ERROR_MESSAGE
+                      ? QString("Errors")
+                      : QString("Warnings");
+  auto itemLine = msgs.begin()->first;
+
+  auto itemMsg =
+      QString("%1 %2 found").arg(QString::number(msgs.size()), itemName);
```
KochynVolodymyr, line 196: 1 Errors found would be a typo

PR #866, opened 2022-12-09
[RG-229] Add simulation clean steps

KochynVolodymyr:
Add simulation clean step which remove wave file.

Tcl command to clean: `simulate <level> clean`
Clean step removes wave file if it exists.

Also Clean task for simulation doesn't clean statuses for downstream tasks because simulation clean doesn't affect on them.

`std::map<SimulationType, std::string> m_waveFiles;` map is used here because need save file for every level.

 - [x] Library: compiler, simulation
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests


ravic-rs:
We need simulation clean checks on upstream codebase.

ravic-rs (approved):
Looks good. We will add tests in upstream since we can't simulate here.

PR #868, opened 2022-12-09
[RG-228] Add close_design tcl command

KochynVolodymyr:
close_design command added.

This command just close current design if it exists. If design doesn't exist, just do nothing.

 - [x] Library: compiler, newproject, foedagcore
 - [ ] Engine
 - [ ] Documentation
 - [x] Regression tests


src/Compiler/Compiler.cpp
```diff
@@ -112,6 +112,7 @@ void Compiler::Help(std::ostream* out) {
   (*out) << "   create_design <name> ?-type <project type>? : Creates a design "
             "with <name> name"
          << std::endl;
+  (*out) << "   close_design     : Close current design" << std::endl;
```
ravic-rs, line 115: Did we verify the behavior when close_design is used from the tcl console in the GUI ?

KochynVolodymyr: yes, I did it

src/MainWindow/main_window.cpp
```diff
@@ -261,9 +261,9 @@ void MainWindow::openProjectDialog(const QString& dir) {
   if (!fileName.isEmpty()) openProject(fileName, false, false);
 }
 
-void MainWindow::closeProject() {
-  if (m_projectManager && m_projectManager->HasDesign() &&
-      confirmCloseProject()) {
+void MainWindow::closeProject(bool force) {
```
ravic-rs, line 264: Why do we need this flag ? close_design could just have been linked to closeProject.

KochynVolodymyr: because there is a message box. We don't need it in tcl command

ravic-rs (approved):

PR #871, opened 2022-12-12
GTKWave: Add install/uninstall targets. Add test to verify install/invoke.

skyler-rs:
> - [x] GTKWave didn't have an install/uninstall target and no test to verify the install was working

>
> - GTKWave now installs to `$(PREFIX)/bin/gtkwave`
>    - Because gtkwave isn't installed under a foedag folder, we have to explicitly uninstall it so the makefile uninstall has been updated to remove `$(PREFIX)/bin/gtkwave`
> - Updated GTKWave cmake to extract after main build to avoid missing folder issues
> - Due to centos CI dependency issues, the main.yml file had to be updated to add gtkwave's lib dir to the LD_LIBRARY_PATH. (This change was made under Nadeem's guidance)
>   - yml required gtk3-devel https://github.com/os-fpga/FOEDAG/pull/871/files#diff-9498e2138e3cedfd5ef27fbf6d93686e3b5d55750938b869a5d15c0c4f6339fdR26
>   - LD_LIBRARY_PATH changes are to allow the test cases in centOs CI to see the gtkwave/lib directory which contains a libtk8.6.so that the rest of the system apparently doesn't have.

> 
> - [x] Library: GTKWave
> - [ ] Engine
> - [ ] Documentation
> - [x] Regression tests
> - [x] Continous Integration (CI) scripts

> - [x] downstream clients will need Makefile and CMake updates to install/uninstall GTKWave in the proper locations


tests/TestGui/gtkwave_invoke.tcl
```diff
@@ -17,23 +17,20 @@
 #You should have received a copy of the GNU General Public License
 #along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-gui_start
+proc delay { time } { set delayDone 0; after $time set delayDone 1; vwait delayDone; }
```
ravic-rs, line 20: Safe to assume this is a blocking call? What's the unit?

ravic-rs (approved):

PR #877, opened 2022-12-14
[RG-233]Introduced timing analysis report for placement/routing tasks

TarasSydor:
Given PR implements RG-233 task and introduces timing report table for Placement/Routing tasks. 

Placement timing report has already been created as a file, however we didn't retrieve actual values but placed entire lines there. After this development actual timing values will be retrieved and put into a structured report (table).
The same is implemented for Routing. Text is slightly different from Placement, but structure is the same and I could reuse the same parsing logic for both, with a tiny tweak.
>
- New reports added to Placement/Routing (in context menu and Reports view): 
image
- Structured report is created:
image
- next to placement, the file is also created for Routing (post_rout_timing.rpt):
image


src/Compiler/CompilerDefines.h
```diff
@@ -105,6 +105,7 @@ static constexpr const char *GLOBAL_PLACEMENT_LOG{"global_placement.rpt"};
 static constexpr const char *PLACEMENT_LOG{"placement.rpt"};
 static constexpr const char *PLACEMENT_TIMING_LOG{"post_place_timing.rpt"};
 static constexpr const char *ROUTING_LOG{"routing.rpt"};
+static constexpr const char *ROUTING_TIMING_LOG{"post_rout_timing.rpt"};
```
skyler-rs, line 108: Typo: e missing in "post_rout**e**_timing.rpt"

src/Compiler/Reports/AbstractReportManager.cpp
```diff
@@ -207,6 +211,36 @@ TaskMessage AbstractReportManager::createWarningErrorItem(
...
+  }
+}
+
+void AbstractReportManager::createTimingDataFile(
+    const QStringList &timingData) {
+  auto logFileName = getTimingLogFileName();
+  if (logFileName.isEmpty()) return;
+
+  auto projectPath = Project::Instance()->projectPath();
+  auto logFilePath = QString("%1/%2").arg(projectPath, logFileName);
```
skyler-rs, line 233: Converting to qstring while creating the path might not be safe on windows.
For these we try to use the std::filesystem so that the / operator will hopefully provide a path that works in both environments.
Example of filesystem path creation: https://github.com/os-fpga/FOEDAG/blob/da0771b6c16f717e0462d16e8fd32e515d0c6835/src/Compiler/Compiler.cpp/#L2242-L2243

skyler-rs (approved):
One filename typo and a path update, but looks good beyond that.

KochynVolodymyr (approved):

ravic-rs (changes requested):
Please address the comments and also include screenshot for Post routing - Report Static timing.

src/Compiler/CompilerDefines.h
TarasSydor: nice catch. fixed

src/Compiler/Reports/AbstractReportManager.cpp
TarasSydor: I just copied that code from another place (I'm an author anyway :) ). fixed

ravic-rs:
@skyler-rs this one's failing for the delay in gtkwave tests.

skyler-rs:
> @skyler-rs this one's failing for the delay in gtkwave tests.

I think if it's still failing even after the delay we should remove this test from the makefile for now. I'm gonna have to look into different ways of making this test more reliable in our CI.

ravic-rs:
> > @skyler-rs this one's failing for the delay in gtkwave tests.
> 
> I think if it's still failing even after the delay we should remove this test from the makefile for now. I'm gonna have to look into different ways of making this test more reliable in our CI.

Please push a PR to disable it temporarily.

skyler-rs:
> Please push a PR to disable it temporarily.

https://github.com/os-fpga/FOEDAG/pull/881

ravic-rs (approved):

PR #882, opened 2022-12-15
EDA-812: if GetNetlistType() is Verilog then pass --ports_info option…

serge-rs:
… to pin_c instead of --blif

EDA-812: if GetNetlistType() is Verilog then pass --ports_info option  to pin_c instead of --blif


serge-rs:
While creating this PR, I encountered a strange merge conflict. I'll close this PR and create another one, to make sure it's not conflicting.

PR #886, opened 2022-12-16
[RG-222] Fix headers not including copyright when launching from install

skyler-rs:
> - [x] To address an existing issue. If so, please provide a link to the issue: RG-222

> - FOEDAG's install of etc had a typo that caused it to install to /usr/local/share/foedag/etc/etc/...
> - FOEDAG's [batch init](https://github.com/os-fpga/FOEDAG/blob/7db2e19a1f1297311f2b092a189ce5da22a90588/src/Main/Foedag.cpp#L364-L374) never set the Context DataPath which is used when looking up the copyright file path.
>    - This would result in all --batch runs not having copyright info in their logs
> - FOEDAG's gui init set the Context DataPath too late ([command stack uses it prior to it being set](https://github.com/os-fpga/FOEDAG/blob/7db2e19a1f1297311f2b092a189ce5da22a90588/src/Main/Foedag.cpp#L184-L189))
>    - This would result in runs of an installed copy of foedag (not executed from build dir) not having copyright info in their logs.
>
> - CMakeLists.txt etc install logic has been updated to properly install to /usr/local/share/foedag/etc/...
> - initBatch now sets the Context DataPath
> - initGui now sets the Context DataPath prior to the CommandStack being created

> 
> - [x] Library: Main
> - [ ] Engine
> - [ ] Documentation
> - [x] Regression tests

> ### Testing
> - Added a test to verify that --batch runs properly add the copyright info to logs.


CMakeLists.txt
```diff
@@ -507,7 +507,7 @@ install(
   DESTINATION cmake)
 
 install(
-  DIRECTORY etc DESTINATION  ${CMAKE_INSTALL_DATAROOTDIR}/foedag/etc/)
+  DIRECTORY etc/ DESTINATION  ${CMAKE_INSTALL_DATAROOTDIR}/foedag/etc/)
```
ravic-rs, line 510: Conflict here with main.

skyler-rs: Resolved, the earlier change to main accomplished the same thing so using that change. Verified it still works w/ `sudo make uninstall` followed by `sudo make install` and checking for duplicate etc folders in /usr/local.

ravic-rs (approved):

PR #890, opened 2022-12-19
[RG-240] Add Timing Analysis settings

skyler-rs:
> - [x] To address an existing issue. If so, please provide a link to the issue: RG-240

> - Timing Analysis engine can only chosen at task execution with the tcl command `sta opensta` which will use OpenSta otherwise, `sta` defaults to tatum/OpenVpr
>
> - Timing Analysis now has task settings for the engine type:
>   image
> - `TimingAnalysisEngineOpt()` has been added to Compiler.cpp
>    - This was necessary to maintain engine choice after a clean as the base TimingAnalysis option blends Clean operations in with engine selection which means when a user does a clean, their engine choice would be lost as well if we use TimingAnalysisOpt

> 
> - [x] Library: Compiler, Main
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

> ### Testing
> - Manually tested in Raptor:
>    - launched with `build/bin/raptor --script test/Testcases/oneff/raptor.tcl`
>    - Let run finish
>    - Double click clean under Timing Analysis
>    - Open Timing Anaylsis settings and change to OpenSTA
>    - Double Click Timing Analysis
>    - R-click Timing Analysis in task list and select `View Timing Analysis Logs`
>      - Ensure OpenSTA was used
>    - Double click clean under Timing Analysis
>    - Open Timing Anaylsis settings and change to Tatum
>    - Double Click Timing Analysis
>    - R-click Timing Analysis in task list and select `View Timing Analysis Logs`
>      - Ensure Tatum/OpenVpr was used


src/Compiler/Compiler.h
```diff
@@ -92,6 +92,7 @@ class Compiler {
   enum class PowerOpt { None, Clean };
   enum class STAOpt { None, Clean, View, Opensta };
```
KochynVolodymyr, line 93: please check if we really need STAOpt::Opensta. It looks like duplication.

src/Compiler/TaskManager.cpp
```diff
@@ -53,6 +53,8 @@ TaskManager::TaskManager(QObject *parent) : QObject{parent} {
                  new Task{"Edit settings...", TaskType::Settings});
   m_tasks.insert(TIMING_SIGN_OFF, new Task{"Timing Analysis"});
   m_tasks.insert(TIMING_SIGN_OFF_CLEAN, new Task{"Clean", TaskType::Clean});
+  m_tasks.insert(TIMING_SIGN_OFF_SETTINGS,
```
KochynVolodymyr, line 56: m_tasks[TIMING_SIGN_OFF]->appendSubTask(m_tasks[TIMING_SIGN_OFF_SETTINGS]); missing

src/Compiler/Compiler.h
skyler-rs: Will do, I was going to ask Alain and Tao to review the new option because currently STAOpt overloads clean and run operations which can result in the Opensta settings getting lost.

src/Compiler/TaskManager.cpp
skyler-rs: Good catch, fixed

skyler-rs:
@taoli4rs & @alain-rs 
STAOpt is currently used for clean operations as well as running w/ OpenVpr or OpenSta. This works because the value is set at tcl command time and persistence isn't required. With settings, we can't store in STAOpt because setting Clean erases the engine choice (opensta or none which is openvpr) when it is set and then the system assumes openVpr after that because clean sets STAOpt to None afterwards.

As such, i've added STAEngineOpt which will just store the Tatum or Opensta choice. This is intended to make as little change as possible so the only change in Compiler logic is we check if either STAOpt has Opensta set (which gets set via tcl when you call `sta opensta`) or STAEngineOpt is set to Opensta. [Related changes to CompilerOpenFPGA.cpp](https://github.com/os-fpga/FOEDAG/pull/890/files#diff-d30b3f4529563e4e488f2f9ea11be3d3f651db9a28fd4d8ca5a83e314653206f)

 This should preserve the existing logic of a tcl command defaulting to openvpr and `sta opensta` running with opensta while allowing the settings system to remember the users last choice when they double click timing analysis to run it, even after a clean.

Could you guys review these specific changes and let me know if this approach is ok?
- [CompilerOpenFPGA.cpp](https://github.com/os-fpga/FOEDAG/pull/890/files#diff-d30b3f4529563e4e488f2f9ea11be3d3f651db9a28fd4d8ca5a83e314653206f)
- [Compiler.h](https://github.com/os-fpga/FOEDAG/pull/890/files#diff-ff9b926a5b5dd44cd6c5cb232a525887cb98ba4b36ef334dfa4d0a2358d8896f)

ravic-rs (approved):
Looks good.

PR #895, opened 2022-12-21
[RG-242] Include histograms into static timing report

TarasSydor:
Given PR finalizes statistical timing reports for Routing/Placement tasks and adds histogram table representation.

The histogram is recognized by its header and is parsed line by line, split into 5 values: From, To, Value, %, Histogram.
(I came up with column names myself, so suggestions\remarks are welcome)

It appears there was no way to indicate the alignment for a specific column (Histogram requires left alignment instead of default center), so the API has been extended to cover it.
>
Statistical reports now look as follows:
image

image


skyler-rs (approved):

KochynVolodymyr (approved):

ravic-rs (approved):
Looks good.

PR #900, opened 2022-12-22
[RG-247] Update log header logic to add header to any file w/ .rpt extension

skyler-rs:
> - [x] To address an existing issue. If so, please provide a link to the issue: RG-247

> - We explicitly add a header to specific .rpt files that we know about during a given task
>   - As a result, unknown .rpt's created by lower level systems were missed. (See link in jira for full break down of missed files and how they were found)
>
> - We now search the project folder for any *.rpt file and add our header to it
>    - Logic has been updated to only add the header if it doesn't already exist
>    - **NOTE:** This currently is non-recursive to protect against the potential case of some sub-task that creates a large number of files in a project sub folder... Not sure if this is a valid concern, but it doesn't hurt for now.
> - Explicit commands to add a header to a given file have been removed from each sub task now that they are dynamically covered.
> - The call to add all headers has been moved into a scope guard within `Compiler::RunCompileTask`
>    - The idea here is that since the main switch case in this function directly returns the results of a call to a given task, we can't call `AddHeadersToLogs` after the task runs unless we add it in a scope guard which will fire once `RunCompileTask` has finished the requested task and then it's scope exits.
>    - By calling `AddHeaderToLogs` here, we will dynamically catch .rpt's created by any task in the switch case even if more are added down the road or sub task report file name are changed etc.

> 
> - [x] Library: Compiler, Utils
> - [ ] Engine
> - [ ] Documentation
> - [x] Regression tests

> ### Testing
> - tests/TestGui/log_header.tcl has been updated to create a fake .rpt file before we run the analyze step and then we verify that this file gets a header added after the analyze step is run.
> - Manually tested in raptor w/ the following steps (see comment linked pr in RG-247 for full context):
>```
> build/bin/raptor --script tests/Testcases/oneff/raptor.tcl
> cd oneff_gemini
> head -n2 *.rpt
>```
>    - Used the output from the commands above to ensure that the previously missed files all have a log header now


ravic-rs (changes requested):
There's a CI problem with test and fake.rpt seems to be missing.

skyler-rs:
> There's a CI problem with test ~and fake.rpt seems to be missing.~

I don't see any failures: https://github.com/os-fpga/FOEDAG/actions/runs/3761705044

ravic-rs:
It went away on a re-run! Seemed like it was hung on a Qt build process.

Get Outlook for iOS<https://aka.ms/o0ukef>

ravic-rs (approved):

PR #907, opened 2022-12-27
[RG-252] Implemented Timing_Analysis parser for reports and messages

TarasSydor:
This task completes RG-252 ticket and introduces parser for Timing Analysis task.

Parser is represented as newly introduced TimingAnalysisReportMessages, that does file parsing and fills Messages with all the reports data. Following reports are available:
- Resource usage;
image

- Circuit statistics; 
image

- Timing report.
image

Reports parsing logic is similar to Routing/Placement tasks, so could be reused. Circuit statistics split was moved to Abstract level, as it's common for Routing and Timing_Analysis.

Messages tab looks as follows:
image


src/Compiler/CompilerDefines.h
```diff
@@ -108,6 +108,7 @@ static constexpr const char *PLACEMENT_TIMING_LOG{"post_place_timing.rpt"};
 static constexpr const char *ROUTING_LOG{"routing.rpt"};
 static constexpr const char *ROUTING_TIMING_LOG{"post_route_timing.rpt"};
 static constexpr const char *TIMING_ANALYSIS_LOG{"timing_analysis.rpt"};
+static constexpr const char *TA_TIMING_LOG{"ta_timing_analysis.rpt"};
```
ravic-rs, line 111: Assuming TA is tatum, why do we have a separate report file for this ? We will either have tatum or sta so it should go into same rpt file.

TarasSydor: Actually TA doesn't stand for Tatum :) it's just there in order not to name the file timing_analysis_timing_analysis.rpt
will try to come up with another name

ravic-rs (approved):

PR #912, opened 2022-12-29
[RG-259] Add timing limits verification for sdc file

KochynVolodymyr:

For the command create_clock added verification of the -period value. This value should be between 0.1 and 1000. Unit is ns.
In case value is out of the range - error is generated:
image

 - [x] Library: compiler
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests


ravic-rs (approved):
The `constexpr auto TimingLimitErrorMessage{"Invalid setting for -period: 0.1 1000"};` needs to abstracted into a generic invalid setting message capable of display both the setting and range as applicable.
We will rework this as part of the logging restructure.

PR #931, opened 2023-01-11
[RG-265] Update clean tasks: remove logs, call downstream clean commands

KochynVolodymyr:
 - [x] To address an existing issue. If so, please provide a link to the issue: RG-265

1. Remove all files from the task when user run Clean. 
2. Run recursively Clean task starting from the last (bitstream generation) till the task requested by user.
3. Cleanup code for `TaskManager::run`. 

New virtual method `GetCleanFiles` should return a list of files which need to remove when Clean is running.
`copyLog` become a static member because it is used in upstream.

 - [x] Library: compiler, utils
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests


src/Compiler/CompilerOpenFPGA.cpp
```diff
@@ -789,6 +789,117 @@ bool CompilerOpenFPGA::DesignChanged(
...
+          "check_rr_node_warnings.log",
+          "packing_pin_util.rpt",
+          "pre_pack.report_timing.setup.rpt",
+          std::string{projectName + "_openfpga.sdc"},
+          std::string{projectName + "_post_synth_ports.json"},
+          "vpr_stdout.log",
+          PACKING_LOG,
+      };
+      break;
+    case Compiler::Action::Detailed:
```
ravic-rs, line 825: Why Detailed ?
```suggestion
    case Compiler::Action::Placement:
```

src/Compiler/CompilerOpenFPGA.cpp
```diff
@@ -789,6 +789,117 @@ bool CompilerOpenFPGA::DesignChanged(
...
+          std::string{projectName + "_post_synth.blif"},
+          std::string{projectName + "_post_synth.edif"},
+          std::string{projectName + "_post_synth.v"},
+          std::string{projectName + "_post_synth.vhd"},
+          std::string{projectName + ".ys"},
+          std::string{projectName + "_synth.log"},
+          SYNTHESIS_LOG,
+      };
+      break;
+    case Compiler::Action::Pack:
```
ravic-rs, line 812:
```suggestion
    case Compiler::Action::Packing:
```

skyler-rs (approved):

ravic-rs:
Note for future: This is a temporary/stop-gap solution to cleaning files for each task. When we get to restructuring the folders and files with a project directory we won't need to hard-code the paths of each files separately to clean.

src/Compiler/CompilerOpenFPGA.cpp
KochynVolodymyr: Here are the changes:
image
Are you sure?

ravic-rs: Ok, let's leave as it is. We should change that to Placement later on. It's an artifact coming from old PR.
https://github.com/os-fpga/FOEDAG/pull/400/files

ravic-rs (approved):

PR #957, opened 2023-01-27
[RG-231] Status icon align to center

KochynVolodymyr:
 - [x] To address an existing issue. If so, please provide a link to the issue: RG-231

Align status icon to the center:
image

 - [x] Library: compiler
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests

### DIFF
--- a/src/Compiler/TaskTableView.cpp
+++ b/src/Compiler/TaskTableView.cpp
@@ -263,13 +263,26 @@ void TaskTableView::TasksDelegate::paint(QPainter *painter,
     if (statusData.type() == QVariant::Bool) {
       label->setMovie(m_inProgressMovie);
       // Place the animation to cells left side, similar to other decorations
-      label->move(m_view.visualRect(index).topLeft());
+      label->move(m_view.visualRect(index).center() -
+                  (label->rect().center() - QPoint{6, 0}));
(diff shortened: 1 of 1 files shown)


src/Compiler/TaskTableView.cpp
```diff
@@ -263,13 +263,26 @@ void TaskTableView::TasksDelegate::paint(QPainter *painter,
...
+      label->move(m_view.visualRect(index).center() -
+                  (label->rect().center() - QPoint{6, 0}));
       m_inProgressMovie->start();
       return;
     } else {
       // Reset the movie when task is no longer in progress
       label->setMovie(nullptr);
     }
+
+    QStyleOptionViewItem opt = option;
```
ravic-rs, line 275: What is this code trying to do ?

KochynVolodymyr: Create new object `opt` and copy all data from object `option`. Copy constructor is used here.

ravic-rs: For alignment, I see the code above but didn't follow why we need this change.

KochynVolodymyr: This code need for alignment. What exactly is not clear?

PR #960, opened 2023-01-31
[RG-237] Save settings when script has finished all tcl commands

KochynVolodymyr:
 - [x] To address an existing issue. If so, please provide a link to the issue: RG-237

Make sure all settings saved after script has finished execution.

 - [x] Library: foedagcore
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests


ravic-rs (approved):
This is a quick-fix for saving settings in script mode. We need to re-write just so that we can separate the settings from the dialog window.
The same will be true for the batch mode which creates project folder etc. but the settings will not be saved.

PR #1022, opened 2023-03-06
[RG-347] Add options for the packing task

KochynVolodymyr:

New options for packing task were added:
image

 - [x] Library: utils, foedagcore
 - [ ] Engine
 - [ ] Documentation
 - [x] Regression tests


etc/settings/settings_test.json
```diff
@@ -257,6 +257,16 @@
 		"default": "Verilog",
 		"arg": "netlist_lang"
 	    },
+	    "allow_unrelated_clustering": {
+                "label": "Allow unrelated clustering",
+                "widgetType": "checkbox",
```
ravic-rs, line 262: By default, both should be "checked" as in "on"

ravic-rs (commented):
Need to set the default state of both checkboxes to checked.

etc/settings/settings_test.json
KochynVolodymyr: This is done in upstream

PR #1038, opened 2023-03-24
Fixed TclWorker batchMode to true

guanyung-rs:

When running batch mode `./Foedag --batch`, the console return immediately instead of entering tcl batch console mode. 
 
Change the TclWorker object instantiation batchMode argument to `true`

Here is the snippet definition of TclWorker constructor code
`TclWorker(TclInterp *interpreter, std::ostream &out,
            std::ostream *err = &std::cerr, bool batchMode = false,
            QObject *parent = nullptr);`
> 
> 
> 
> 
> 
>
> 
> 
> 
> 
> 

> 
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests


alain-rs:
@guanyung-rs  can you restart the Action, they are hanging.
This might be a problem for foedag --batch --script test.tcl
We want that to exit even if the test does not contain an exit command.
You might have to check if a --script  or a --cmd has been passed, then exit after they are executed.
You only want to stay in the interpreter if there is no script: foedag --batch

PR #1052, opened 2023-04-05
[RG-360] Change synthesis settings for bram and dsp

KochynVolodymyr:

2 new synthesis settings were added for DSP and BRAM:
image

Previous settings removed.

 - [x] Library: compiler, foedagcore, utils
 - [ ] Engine
 - [ ] Documentation
 - [x] Regression tests

### DIFF
--- a/etc/settings/settings_test.json
+++ b/etc/settings/settings_test.json
@@ -217,15 +217,23 @@
                 "default": "Verilog",
                 "arg": "netlist_lang"
             },
-            "no_dsp_checkbox": {
-                "label": "No DSP Blocks",
-                "widgetType": "checkbox",
-                "arg": "no_dsp"
+            "dsp_spinbox_ex": {
+                "label": "Dsp limit",
+                "widgetType": "Spinbox",
+                "minVal": 0,
+                "maxVal": 0,
+                "stepVal": 1,
+                "default": 0,
+                "arg": "dsp_limit"
             },
-            "no_bram_checkbox": {
-                "label": "No Block Ram",
-                "widgetType": "checkbox",
-                "arg": "no_bram"
+            "bram_spinbox_ex": {
+                "label": "Block Ram limit",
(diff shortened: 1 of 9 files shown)


ravic-rs:
@KochynVolodymyr macos-clang failed.

etc/settings/settings_test.json
```diff
@@ -217,15 +217,23 @@
...
-            },
-            "no_bram_checkbox": {
-                "label": "No Block Ram",
-                "widgetType": "checkbox",
-                "arg": "no_bram"
+            "dsp_spinbox_ex": {
+                "label": "Dsp limit",
+                "widgetType": "Spinbox",
+                "minVal": 0,
+                "maxVal": 154,
```
ravic-rs, line 224: I believe this value is dependent on the device.xml so we can't hardcode it.

ravic-rs: Plus it will be dependent on the device user selects for a design project.

KochynVolodymyr: I've put zeros.

alaindargelas: That hardcoded number "154" is just as an example right? The actual value needs to be read from the device.xml, the number is different for every devices.

alaindargelas: Also, by default is has to be set to the max range, not 0

KochynVolodymyr: this value will be updated after device.xml loaded.

KochynVolodymyr: > Also, by default is has to be set to the max range, not 0

@alaindargelas different devices have this value different. What max range do you mean? anyway we have another json upstream

ravic-rs: We reload the settings dependent on device.xml using the reloadSettings() function. This is a new feature since now we might have values in settings_test.json that is dependent on device.xml.

ravic-rs (approved):

PR #1135, opened 2023-06-01
[EDA-1607] Add dumpfile for icarus

KochynVolodymyr:
 - [x] To address an existing issue. If so, please provide a link to the issue: EDA-1607

Dump waveform file with user filename for icarus

 - [x] Library: simulation
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests


src/Simulation/Simulator.cpp
```diff
@@ -735,6 +735,7 @@ std::string Simulator::SimulatorRunCommand(SimulationType simulation,
       if (m_waveType == WaveformType::FST) {
         command += " -fst";
       }
+      if (!m_waveFile.empty()) command += " -dumpfile=" + m_waveFile;
```
ravic-rs, line 738: What's the default name for wavefile in our GUI ? It appears to be "icarus" which doesn't align with other simulator defaults, correct ?

ravic-rs: Based off the settings, I think it should have been "sim_rtl.fst"
We only support fst for icarus via GUI as per the simulator docs.
image

KochynVolodymyr: this is issue with simulate command, I'll create another PR

KochynVolodymyr: Here is fix: https://github.com/os-fpga/FOEDAG/pull/1138

PR #1147, opened 2023-06-15
Changing path of json file containing ports information

behzadmehmood-rs:

>  Currently the json file for ports information is created in the design directory.
>
> This PR changes its path to the Project directory.

> 
> - [x] Engine
> - [ ] Documentation
> - [ ] Regression tests

> - [x] Break back-compatibility. If so, please list who may be influenced.
> Needs synchronisation with `Raptor_Tools` changings.


src/Compiler/CompilerOpenFPGA.cpp
```diff
@@ -800,7 +800,7 @@ std::vector<std::string> CompilerOpenFPGA::GetCleanFiles(
                std::string{topModule + "_post_synthesis.sdf"},
                std::string{topModule + "_post_synthesis.v"},
                std::string{projectName + "_sta.cmd"},
-               std::string{projectName + "_post_synth_ports.json"},
+               "post_synth_ports.json",
```
ravic-rs, line 803: Why are we hard-coding the name of json here  and other places ?

ravic-rs (changes requested):
Need details on change in json naming format.

src/Compiler/CompilerOpenFPGA.cpp
behzadmehmood-rs: This file is generated from the synthesized netlist so we can name it "post_synth_ports.json". The reason for hardcoding the name is that for `Post-synthesis Project`, instead of being generated as `projectName + "_post_synth_ports.json"` it is being generated with the name of the design (netlist) file, and cannot be cleaned using `projectName + "_post_synth_ports.json"`. Same is the case with some other files like `.place` and `.route`, these are not being cleaned properly for `Post-synthesis Project`. I have resolved the issue for ports file, kindly resolve it for others too. 
Kindly give your suggestion if there is some better way.

ravic-rs: Thanks for letting me know. We are changing the folder structure and the issues you noticed should be fixed then.
@shivaahir1508 just FYI that we should be testing both RTL and Post-synthesis project types because of these kind of differences.

ravic-rs (approved):

PR #1181, opened 2023-07-18
[EDA-1628] Change _post_synthesis* to _post_route*

ravic-rs:
> - [X] To address an existing issue. If so, please provide a link to the issue: EDA-1628

> Renamed files to match the compilation tasks

> 
> 
> 
> 
> 

> 
> - [X] Library: Compiler and Simulation
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests
> - [ ] Continuous Integration (CI) scripts

Will require backend updates to reflect the naming change.
> - [X] Break back-compatibility. If so, please list who may be influenced.
> All previous foedag projects.


alain-rs (commented):
That will not work as-is.
You have to actually "move" as in "operating system move" the files post routing.
Here you are changing the name of the files when reading. You have to actually change the names of the files created post routing too.

ravic-rs:
> That will not work as-is. You have to actually "move" as in "operating system move" the files post routing. Here you are changing the name of the files when reading. You have to actually change the names of the files created post routing too.

Thanks for looking at this.
image

From the image above,
Are we moving the files within the timing analysis directory after it has completed? What if the TA fails?
It appears that these files are generated by backend so it's not clear when we should "mv" these plus the case where user uses OpenSTA.

alaindargelas:
- Under the routing directory, you can move within the routing directory in the routing stage itself
- Under the timing analysis directory, I think you can do the same at the correct step in the "analyze" step, you can check with @serge-rs

PR #1186, opened 2023-07-20
Remove bitstream cache

chungshien:
> - [x] Breaking new feature. If so, please describe details in the description part.

This PR remove bitstream cache support that we previously add, for reasons:
a. We have better solution to generate latch bitstream in fast and low memory footprint
b. The cache files are incorrect/buggy in the first place
c. Generating/reading the cache take long time and high memory usage

Beside that, make two additional enhancement:
a. Allow user to set the WL ordering (by default to be incremental, can use option to change to decremental format - original openfpga generation format) - validation team want it to be incremental. 
b. Allow user to ignore don't care bit (by default don't care bit remained as 'x', can use option to change it to '0')


ecchg (approved):

chungshien:
@alain-rs 
@ravic-rs is the codecov only cover the the one in tests/unittest? 

I believe my change is very hard to test through unit test. Compiler/CompilerOpenFPGA need a lot of setup TCL, project, UI, session etc. 

I try to test my change through tests/TestGui and tests/Testcases, but base on the coverage script, it does not cover those (maybe it cannot cover as well)

Or you have suggestion how to cover code change in Compiler/CompilerOpenFPGA?

alaindargelas:
@chungshien , you can checkin. No coverage really necessary for this.

ravic-rs:
> @alain-rs @ravic-rs is the codecov only cover the the one in tests/unittest?
> 
> I believe my change is very hard to test through unit test. Compiler/CompilerOpenFPGA need a lot of setup TCL Interp, project, UI, session etc.
> 
> I try to test my change through tests/TestGui and tests/Testcases, but base on the coverage script, it does not cover those (maybe it cannot cover as well)
> 
> Or you have suggestion how to cover code change in Compiler/CompilerOpenFPGA?

Codecov checks are placeholders and we know that it's not ready. We might disable it temporarily, for now as Alain mentioned you can check-in.
It's encouraged to add gtests where you can.

chungshien:
I had been spending a lot of hours on this. 
If it is not ready, then we should not enable that to waste engineer time, or at least customize the script to skip/bypass certain subsystem(s) that is impossible to unit test because the way it is coded.

PR #1192, opened 2023-07-25
[EDA-957] Initial implementation of Compute Usage  table which shows task duration and memory utilization

KochynVolodymyr:

Add new feature - Compute Usage:
image

Result is available only after task successfully done.
 
Result is N/A in case:
* Task was cleaned
* Task was skipped because of no design change.
* Project was opened.

For simulation task:
* Duration is a sum of three (elaboration, compile and simulate)
* Utilization is maximum of three  (elaboration, compile and simulate)

 - [x] Library: compiler, foedagcore
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests

### DIFF
--- a/src/Compiler/Compiler.cpp
+++ b/src/Compiler/Compiler.cpp
@@ -2101,11 +2101,13 @@ bool Compiler::Compile(Action action) {
   if (task != TaskManager::invalid_id && m_taskManager) {
     m_taskManager->task(task)->setStatus(TaskStatus::InProgress);
   }
+  m_utils = {};
   res = SwitchCompileContext(
       action, [this, action]() { return RunCompileTask(action); });
   if (task != TaskManager::invalid_id && m_taskManager) {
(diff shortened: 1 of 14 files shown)


ravic-rs:
image

ravic-rs:
Add "Why we don't have this for simulation tasks" to the description.

alain-rs:
@ravic-rs This should not be called "Performance", it is misleading. The Performance of an FPGA is its Fmax.
This should be called "Compute usage" or something like that

ravic-rs:
> @ravic-rs This should not be called "Performance", it is misleading. The Performance of an FPGA is its Fmax. This should be called "Compute usage" or something like that

@alain-rs the term's coming from "perf.log". Is "Compilation Performance" better? Let us know else we will go with "Compute Usage".

alain-rs:
Let's go for Compute usage.

PR #1201, opened 2023-07-31
No assertion abort

chungshien:
> - [x] To address an existing issue. Issue: https://rapidsilicon.atlassian.net/browse/EDA-1829

> 
> 
> 
> 
> 
>
> 
> 
> 
> 
> 

> 
> - [x] Engine
> - [ ] Documentation
> - [ ] Regression tests


alain-rs (changes requested):
You coded the try catch in the unit test only.
You need to add try catch block at the highest level of your component API,
if you don't do that, you are throwing an exception but no one catches it and the GUI will exit.

chungshien:
Configuration subsystem is architected that there is only one single point of entry for outside world to call configuration core function via CFGCompiler.

This is screenshot of the spec
image

Hence only need to change one place to use try-and-catch, which is CFGCompiler::Configure()

alain-rs (approved):

PR #1228, opened 2023-08-18
Programmer API library implementation

guanyung-rs:
> - [x] Breaking new feature. If so, please describe details in the description part.

Implemented Programmer Library APIs
`int InitLibrary(std::string openOCDPath)`
`std::string GetErrorMessage(int errorCode)`
`int GetAvailableCables(std::vector<Cable>& cables)`
`int ListDevices(const Cable& cable, std::vector<Device>& devices)`
`int GetFpgaStatus(const Cable& cable, const Device& device, CfgStatus& status)`
`int ProgramFpga(...)`
`int ProgramFlash(...)`

Implemented a bunch of helper functions and also added the unit tests for them

Added libusb library dependencies. The programmer library API is using libusb APIs to detect the connected USB-JTAG cables

Make changes to CI build main.yml and a couple of CMakefiles so it downloads and installs the necessary packages for libusb library dependencies for build and installation.

#### Testing
- There are unit tests to cover the helper functions
- The Programmer API is tested on the emulator system with ubuntu 20.04 OS
> 
> 
> 
> 
> 
>
> 
> 
> 
> 
> 

> 
> - [ ] Engine
> - [ ] Documentation
> - [ ] Regression tests

### DIFF
--- a/src/Configuration/CFGCommon/CFGCommon.cpp
+++ b/src/Configuration/CFGCommon/CFGCommon.cpp
@@ -422,10 +422,8 @@ int CFG_execute_cmd(const std::string& cmd, std::string& output) {
   return exitcode;
 }
 
-int CFG_execute_cmd_with_callback(
-    const std::string& cmd, std::string& output, std::ostream* outStream,
-    std::regex patternToMatch, std::atomic<bool>& stopCommand,
-    std::function<void(const std::string&)> callback) {
+int CFG_execute_cmd(const std::string& cmd, std::string& output,
+                    std::ostream* outStream, std::atomic<bool>& stopCommand) {
 #ifdef _WIN32
 #define POPEN _popen
 #define PCLOSE _pclose
@@ -453,13 +451,55 @@ int CFG_execute_cmd_with_callback(
     if (stopCommand) {
       break;
     }
+  }
+
+  int status = PCLOSE(pipe);
+  int exit_code = WEXITSTATUS(status);
+  return exit_code;
+}
+
+int CFG_execute_cmd_with_callback(
+    const std::string& cmd, std::string& output, std::ostream* outStream,
+    std::regex patternToMatch, std::atomic<bool>& stopCommand,
+    std::function<void(const std::string&)> progressCallback,
+    std::function<void(const std::string&)> generalCallback) {
+#ifdef _WIN32
+#define POPEN _popen
+#define PCLOSE _pclose
+#else
+#define POPEN popen
+#define PCLOSE pclose
+#endif
+
+  FILE* pipe = POPEN(cmd.c_str(), "r");
+  if (pipe == nullptr) {
+    return -1;
+  }
+
+  // Read the output of the command and store it in the output string.
+  char buffer[1024];
+  std::string newline;
+  while (fgets(buffer, sizeof(buffer), pipe) != nullptr && !stopCommand) {
+    output += buffer;
+    newline = buffer;
+    if (generalCallback != nullptr) {
+      generalCallback(newline);
+    }
+
+    if (outStream) {
+      *outStream << newline;
+    }
+
+    if (stopCommand) {
+      break;
+    }
 
     std::smatch matches;
     if (std::regex_search(newline, matches, patternToMatch)) {
-      if (callback != nullptr) {
+      if (progressCallback != nullptr) {
         std::string strOutput = matches.str();
         strOutput += "\n";
-        callback(strOutput);
+        progressCallback(strOutput);
       }
     }
   }
--- a/src/Configuration/Programmer/Programmer_cmd.cpp
+++ b/src/Configuration/Programmer/Programmer_cmd.cpp
@@ -1,9 +1,9 @@
 /*
-Copyright 2021 The Foedag team
+Copyright 2023 The Foedag team
 
 GPL License
 
-Copyright (c) 2021 The Open-Source FPGA Foundation
+Copyright (c) 2023 The Open-Source FPGA Foundation
 
 This program is free software: you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
@@ -19,8 +19,14 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <iostream>
+
 #include "CFGCommon/CFGCommon.h"
 #include "Programmer.h"
+
+void progressCallback(std::string msg) { std::cout << msg; }
+
+void messageCallback(std::string msg) { std::cout << "MS-CALLBACK :" << msg; }
 
 using namespace FOEDAG;
 int main(int argc, const char** argv) {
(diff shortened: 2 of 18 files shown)


ravic-rs:
@guanyung-rs if this is ready, please create corresponding PRs at upstream repos. My understanding is that we will need to update CI scripts there too.

ravic-rs:
> @guanyung-rs if this is ready, please create corresponding PRs at upstream repos. My understanding is that we will need to update CI scripts there too.

As mentioned in previous PR, for any build changes please seek review from @nadeemyaseen-rs as we need to plan for bundling new packages for deployment.

guanyung-rs:
> @guanyung-rs if this is ready, please create corresponding PRs at upstream repos. My understanding is that we will need to update CI scripts there too.

FYI, I created PRs for the upstream repos

src/Configuration/Programmer/Programmer_cmd.cpp
```diff
@@ -19,57 +19,155 @@ You should have received a copy of the GNU General Public License
...
+  CFG_POST_MSG("### ListDevices API testing ###");
+  for (auto& device : devices) {
+    CFG_POST_MSG("-------------------------");
+    CFG_POST_MSG("Device name: %s", device.name.c_str());
+    CFG_POST_MSG("Device index: %d", device.index);
+    CFG_POST_MSG("Device jtagId: 0x%08x", device.tapInfo.idCode);
+    CFG_POST_MSG("Device mask: 0x%08x", device.tapInfo.irMask);
+    CFG_POST_MSG("Device irlength: %d", device.tapInfo.irLen);
+    CFG_POST_MSG("Device flashSize: %s bits",
+                 formatMemorySize(device.flashSize).c_str());
```
chungshien, line 120: Any where in the code do the checking of flashsize validity? For example, it must be multiple 8bits, powerof2. Without this checking before calling formatMemorySize() conversion, then the conversion is wrong.

src/Configuration/Programmer/Programmer_cmd.cpp
```diff
@@ -19,57 +19,155 @@ You should have received a copy of the GNU General Public License
...
+    CFG_POST_MSG("channel   : %d", cable.channel);
+  }
+
+  CFG_POST_MSG("-- ListDevices(cables, outputMsg) --")
+  ret = ListDevices(cables[0], devices);
+  if (ret != NoError) {
+    CFG_POST_MSG("Error code = %d. %s", ret, GetErrorMessage(ret).c_str());
+    return ret;
+  }
+  if (devices.size() <= 0) {
```
chungshien, line 107: strictly speaking the checking should be "=="

size() for vector/list is size_t which is an unsigned number. neg is invalid.

src/Configuration/Programmer/Programmer_cmd.cpp
guanyung-rs: The flash size value is read from the return of openOCD. I would think that the validity of the flash size is checked in openOCD level.

The formatMemorySize is a lambda from line 43 to 54. There is no checking in this lambda function as its name implied it just do the formatting.

```auto formatMemorySize = [](int sizeInBits) -> std::string {
    const int bitsInKilobit = 1024;
    const int bitsInMegabit = 1024 * 1024;

    if (sizeInBits >= bitsInMegabit) {
      int megabits = sizeInBits / bitsInMegabit;
      return std::to_string(megabits) + "M";
    } else if (sizeInBits >= bitsInKilobit) {
      int kilobits = sizeInBits / bitsInKilobit;
      return std::to_string(kilobits) + "K";
    } else {
      return std::to_string(sizeInBits);
    }
  };

I can add the checking of flash size value before the formatting happen.

Comment (chungshien) [reply src/Configuration/Programmer/Programmer_cmd.cpp:120]: The checking should be up-front. It should be done in the API that read devices from openocd (which should be the entry point to any use case), not during the formatting. The validation should be in place for any use case even maybe some use case does not need to display or format.

Comment (chungshien) [reply src/Configuration/Programmer/Programmer_cmd.cpp:120]: I would like to see, when we rely on other subsystem/module/APIs to get some data, we should always validate the correctness of the data, before we use it.

This should be applicable for general SW development, not only flashsize in this case.

Comment (guanyung-rs) [reply src/Configuration/Programmer/Programmer_cmd.cpp:120]: Ok. I will add the flash size value check during the ListDevice API entry point.

File: src/Configuration/Programmer/Programmer_cmd.cpp (hunk 2 above)
Comment (guanyung-rs) [reply src/Configuration/Programmer/Programmer_cmd.cpp:107]: Ok. I will fix it to use '=='

File: src/Configuration/Programmer/Programmer_cmd.cpp [hunk 3]
```diff
@@ -19,57 +19,155 @@ You should have received a copy of the GNU General Public License
...
-    CFG_POST_MSG("### ProgressCallback API testing ###");
-    CFG_POST_MSG("ProgressCallback percent: %s",
-                 std::to_string(percent).c_str());
-  };
+  int userInput = 1;
   std::atomic<bool> stop = false;
-  ProgramFpga(device, "test.bit", "config.cfg", nullptr, callback, progress,
-              stop);
+  std::string bitfile =
+      "/home/asic01/development/ubi_fsbl_fpga_faked_spare_reg.bin";
```
KochynVolodymyr, line 137: Is that expected this file to be hardcoded?

src/Configuration/Programmer/Programmer_cmd.cpp
```diff
@@ -19,57 +19,155 @@ You should have received a copy of the GNU General Public License
...
+  std::cin >> userInput;
+
+  if (userInput == 1) {
+    ret = ProgramFpga(cables[0], devices[0], bitfile, stop,
+                      nullptr,          //&std::cout,
+                      nullptr,          // messageCallback, //nullptr
+                      progressCallback  // progressCallback
+    );
+  } else {
+    bitfile = "/home/asic01/development/ubi_fsbl_fpga_flash_program.bin";
```
KochynVolodymyr, line 149: Is that expected this file to be hardcoded?

src/Configuration/Programmer/Programmer_helper.cpp
```diff
@@ -0,0 +1,453 @@
...
+static const std::vector<std::string> programmer_subcmd{
+    "fpga_config", "fpga_status", "flash", "list_device", "list_cable"};
+
+std::string transportTypeToString(TransportType transport) {
+  switch (transport) {
+    case TransportType::jtag:
+      return "jtag";
+      // Handle other transport types as needed
+  }
+  return "";
```
KochynVolodymyr, line 42: Please use std::string{} instead of `""`

src/Configuration/Programmer/Programmer_helper.cpp
```diff
@@ -0,0 +1,453 @@
...
+"2  omap5912.unknown      Y    0x00000000 0x00000000 8     0x01  0x03\n"
+"3 auto0.tap              Y     0x20000913 0x00000000     5 0x01  0x03
+^\d+\s+\w+\.\w+\s+\w+\s+0x[\da-fA-F]+\s+0x[\da-fA-F]+\s+\d+\s+0x[\da-fA-F]+\s+0x[\da-fA-F]+$
+*/
+std::vector<TapInfo> extractTapInfoList(const std::string& tapInfoString) {
+  std::vector<TapInfo> tapInfoList;
+  std::istringstream iss(tapInfoString);
+  std::string line;
+  std::string token;
+  size_t pos = -1;
```
KochynVolodymyr, line 119: size_t is unsigned type. Looks suspicious

src/Configuration/Programmer/Programmer_helper.cpp
```diff
@@ -0,0 +1,453 @@
...
+         Device               ID           IRLen      Flash Size
+-------- -------------------- ------------ ---------- ----------------
+Found  0 Gemini               0x20000913   5          16384
+*/
+std::vector<Device> extractDeviceList(const std::string& devicesString) {
+  std::vector<Device> devices;
+
+  std::istringstream iss(devicesString);
+  std::string line;
+  size_t pos = -1;
```
KochynVolodymyr, line 171: size_t is unsigned type. Looks suspicious

src/Configuration/Programmer/Programmer_helper.cpp
```diff
@@ -0,0 +1,453 @@
...
+    std::getline(lineStream, indexStr, ' ');  // index string
+    device.index = static_cast<int>(CFG_convert_string_to_u64(indexStr));
+
+    std::getline(lineStream >> std::ws, device.name, ' ');  // device name
+
+    std::string idCodeStr;
+    std::getline(lineStream >> std::ws, idCodeStr,
+                 ' ');  // idCode string in hex
+    device.tapInfo.idCode =
+        static_cast<uint32_t>(CFG_convert_string_to_u64(idCodeStr));
```
KochynVolodymyr, line 200: Is that expected we might lose data due to all conversions like this ?

src/Configuration/Programmer/Programmer_helper.cpp
```diff
@@ -0,0 +1,453 @@
...
+std::string buildScanChainCommand(const Cable& cable) {
+  std::stringstream cmd;
+  cmd << buildFpgaCableCommand(cable, "scan_chain");
+  return cmd.str();
+}
+
+std::string buildListDeviceCommand(const Cable& cable,
+                                   const std::vector<TapInfo>& foundTapList) {
+  if (foundTapList.size() == 0) {
+    return "";
```
KochynVolodymyr, line 335: return std::string{};

src/Configuration/Programmer/Programmer_cmd.cpp
guanyung-rs: This programmer_cmd is an executable for manual testing and also serve as an example for using the APIs. The hardcoded file is the test bit file for programming. User should their own bitstream file generated

src/Configuration/Programmer/Programmer_cmd.cpp
guanyung-rs: see above reply.

src/Configuration/Programmer/Programmer_helper.cpp
guanyung-rs: Yes. I did this with purpose to initialize the pos variable to -1. It will end up as a very large unsigned data as a invalid number. A zero may give some meaningful value.

src/Configuration/Programmer/Programmer_helper.cpp
guanyung-rs: the idCode is unsigned 32bit. So the data lose doesn't matter.

src/Configuration/Programmer/Programmer_helper.cpp
chungshien: GY, you should change this to std::string::npos. They are equivalent, but using npos is more obvious -- to initiailize with invalid pos.

guanyung-rs: Noted. Make sense.

src/Configuration/Programmer/Programmer_helper.cpp
```diff
@@ -0,0 +1,453 @@
...
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "Programmer_helper.h"
+
+#include <regex>
+#include <unordered_set>
+
+#include "../../src/Utils/StringUtils.h"
```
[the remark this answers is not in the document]
guanyung-rs: Fixed. It is now `#include "Utils/StringUtils.h"`

src/Configuration/Programmer/Programmer_helper.cpp
guanyung-rs: Fixed. Change to std::string:npos

src/Configuration/Programmer/Programmer_helper.cpp
```diff
@@ -0,0 +1,453 @@
...
+}
+
+CfgStatus extractStatus(const std::string& statusString, bool& statusFound) {
+  const std::string pattern(R"(\s*(\d+)\s+(\S+)\s+(\d+)\s+(\d+)\s*)");
+  CfgStatus status;
+  status.cfgDone = false;
+  status.cfgError = false;
+  std::istringstream iss(statusString);
+  std::string line;
+  size_t pos = -1;
```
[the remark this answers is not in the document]
guanyung-rs: Fixed. Change to std::string:npos

src/Configuration/Programmer/Programmer_helper.cpp
```diff
@@ -0,0 +1,453 @@
...
+  }
+
+  return status;
+}
+
+std::stringstream buildFpgaCableStringStream(const Cable& cable) {
+  std::stringstream ss;
+  ss << std::hex << std::showbase;
+  ss << " -c \"adapter driver ftdi\"";
+  if (cable.serialNumber != "") {
```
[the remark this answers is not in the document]
guanyung-rs: Fixed.

src/Configuration/Programmer/Programmer_helper.cpp
guanyung-rs: Fixed. Change to std::string:npos

src/Configuration/Programmer/Programmer_cmd.cpp
guanyung-rs: Rename to `bitfile = "/your/flash/bitstream/file.bin";`

src/Configuration/Programmer/Programmer_cmd.cpp
guanyung-rs: @chungshien I added 
- flash size checking [here](https://github.com/os-fpga/FOEDAG/blob/00f2b400a42fc0af4ca32e8fdc654d636a2ccc4e/src/Configuration/Programmer/Programmer_helper.cpp#L214C1-L216C6)
- the number to unit string conversion [function ](https://github.com/os-fpga/FOEDAG/blob/00f2b400a42fc0af4ca32e8fdc654d636a2ccc4e/src/Configuration/CFGCommon/CFGCommon.cpp#L319C1-L328C2)in cfgcommon 
- added the unit tests for `CFG_convert_number_to_unit_string`

src/Configuration/CFGCommon/CFGCommon.cpp
```diff
@@ -316,6 +316,17 @@ uint64_t CFG_convert_string_to_u64(std::string string, bool no_empty,
   return value;
 }
 
+std::string CFG_convert_number_to_unit_string(uint64_t number) {
+  static const std::vector<std::string> units = {"", "K", "M", "G", "T"};
+  size_t unitIndex = 0;
+  while (number >= 1024 && unitIndex < units.size() - 1) {
```
chungshien, line 322: Please use the original checking that I gave you.

You must check (% 1024) == 0.

chungshien: If the number is 1500. Then we should not convert it (because the result will be 1k which is wrong) in this case unless we support float number. 

I know you did check (% 1024) == 0 in the extractDeviceList() [by the way it is wrong checking]. But remember this is a generic number, it is not used by flashsize conversion.

src/Configuration/Programmer/Programmer_helper.cpp
```diff
@@ -0,0 +1,453 @@
...
+    device.tapInfo.irLen =
+        static_cast<uint32_t>(CFG_convert_string_to_u64(irLenStr));
+
+    std::string flashSizeStr;
+    std::getline(lineStream >> std::ws, flashSizeStr,
+                 ' ');  // flash size string in base 10
+    device.flashSize =
+        static_cast<uint32_t>(CFG_convert_string_to_u64(flashSizeStr));
+    // check flashSize is multiple of 1024
+    if (device.flashSize % 1024 != 0) {
```
chungshien, line 214: As I mentioned in original comment, the checking of flash size should be
1. multiple of 8 (because it is in bits, we do not have 6bits memory for example) 
2. powerof2 which mean it should be 8, 16, 32, 64, 128, 1024, 2048, 4096, 8k, 16k, 32k, 1M, 2M, 4M, 8M, 128M, 1G, 512G etc

If the flashsize is 3*1024 = 3072, which can be divided by 1024 but it is still an invalid size. I do not think you can buy embedded flash with 7M, 800M size.

src/Configuration/CFGCommon/CFGCommon.cpp
guanyung-rs: Ok. Fixed.

src/Configuration/Programmer/Programmer_helper.cpp
guanyung-rs: Fixed.

src/Configuration/Programmer/Programmer_cmd.cpp
```diff
@@ -19,57 +19,140 @@ You should have received a copy of the GNU General Public License
...
-  ProgressCallback progress = [](double percent) {
-    CFG_POST_MSG("### ProgressCallback API testing ###");
-    CFG_POST_MSG("ProgressCallback percent: %s",
-                 std::to_string(percent).c_str());
-  };
+  int userInput = 1;
   std::atomic<bool> stop = false;
-  ProgramFpga(device, "test.bit", "config.cfg", nullptr, callback, progress,
-              stop);
+  std::string bitfile = "/your/bitstream/file.bin";
```
KochynVolodymyr, line 121: I don't understand this. Is that fake path and we don't care about real path? Why isn't it empty? 
If it matters it should be real path that could be read from some configuration file, for example. We can't push it like this to public repo.

chungshien (approved):

src/Configuration/Programmer/Programmer_cmd.cpp
guanyung-rs: offline discussion to resolve this question

esther-rs (approved):

PR #1284, opened 2023-09-11
Update main.yml to add if always to upload artifact

nadeemyaseen-rs:
Always upload the artifacts.


KochynVolodymyr (approved):

ravic-rs:
@nadeemyaseen-rs why do we need this ?

nadeemyaseen-rs:
In case when make is failing, we have log files as artefacts to do the debug.


Regards

Nadeem Yaseen

PR #1295, opened 2023-09-14
safe hierarchy building

alain-rs:
No matter how incomplete the json is, the hier builder should not crash.


KochynVolodymyr:
@alain-rs you reverted my changes

PR #1288, opened 2023-09-13
Add tcl console to programmer

KochynVolodymyr:
### Motivate of the pull request
 - [ ] To address an existing issue. If so, please provide a link to the issue: <issue id>
 - [x] Breaking new feature. If so, please describe details in the description part.

 ### Describe the technical details
Add tcl console support in programmer 
image

 ### Which part of the code base require a change
 
 - [x] Library: programmer-gui
 - [ ] Plug-in: <Specify the plugin name>
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests
 - [ ] Continous Integration (CI) scripts

PR #1307, opened 2023-09-20
Restore the broken code from PR:1288

chungshien:

Existing flow is broken by https://github.com/os-fpga/FOEDAG/pull/1288

1. Ravi and Volodymyr think that Programmer is a standalone app and they thought project information is not needed by configuration
2. They remove project setting in configuration subsystem via https://github.com/os-fpga/FOEDAG/pull/1288
3. This is only true for Programmer (used to program bitsstream)
4. However configuration is not all JUST about Programmer, it involves a lot of project related feature: OCLA, bitstream generation, PGM database that tied to device, and a lot more future feature: Partial Reconfiguration, SEU, etc
5. The correct way to do it is
>>> If there is project, then we should set the information for other configuration feature to proceed
>>> If there is no project, then this is standalone app (Programmer GUI), we can skip setting those info. 

Other additional change is dummy change so that we can add unit test to hit the code coverage requirement - please be reminded that it is very hard to unit test Compiler class because it involves a lot of class that is very hard to initialize in unit test. 

> 
> - [x] Engine
> - [ ] Documentation
> - [ ] Regression tests


ravic-rs:
@chungshien my understanding is that both programmer and debugger are supposed to be used in standalone mode. That was the reason as you know to remove access to the project manager in configuration module after agreeing with @guanyung-rs .

The following information was not known then. I think there should be a better way to switch the flow than to check for project manager existence.
_"However configuration is not all JUST about Programmer, it involves a lot of project related feature: OCLA, bitstream generation, PGM database that tied to device, and a lot more future feature: Partial Reconfiguration, SEU, etc"_

esther-rs (approved):

chungshien:
@ravic-rs , this is just a very straightforward checking that if we want to access something from a class, make sure this class is not null (and it make perfect sense since in this use case there is possible it might be null). I don't think this checking has any issue or cannot last long and need "a better way".

Other function in Compiler has done such null checking before proceeding as well: https://github.com/os-fpga/FOEDAG/blob/main/src/Compiler/Compiler.cpp#L2025-L2026

@alain-rs , can you help to merge this PR, I do not think this particular change can easily hit the required coverage, as it involve Compiler which is not unit-test-friendly

alain-rs:
@chungshien , this branch is out of date, please sync first

PR #1322, opened 2023-09-27
raptor: use eblif for vhdl flow

serge-rs:
This allows pin_c and bitstream generation to work when Raptor uses VHDL input


ravic-rs:
@KochynVolodymyr FYI: Since we don't have verific, we cant use vhdl netlist, workaround solution till we enable it back.

PR #1353, opened 2023-10-20
[RG-486] When --project switch is used, open specified project but do not start compilation

KochynVolodymyr:
 - [x] To address an existing issue. If so, please provide a link to the issue: RG-486

We do not run project when `--project` switch is used. 

 - [x] Library: foedagcore
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests


ravic-rs (approved):
I think we should update the tcl cmd help to reflect this change.

KochynVolodymyr:
@ravic-rs Here is a description
image

It looks like we don't need to update help

PR #1369, opened 2023-11-08
Fixed tcl simple parser to calculate commands correct

KochynVolodymyr:

Fixed command count for tcl script. 
This count reflect here: 
image

 - [x] Library: foedagcore
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests


ravic-rs (approved):

ravic-rs (changes requested):
We are only counting tasks from tcl script which are in the GUI table.
I don't think we need this.
Also there's no statues for non-tasks tcl commands.

KochynVolodymyr:
> We are only counting tasks from tcl script which are in the GUI table. I don't think we need this. Also there's no statues for non-tasks tcl commands.

@ravic-rs That's correct. I just fixed the way how it is done now. For now, it calculates it in a wrong way.

PR #1371, opened 2023-11-09
Qt6 review

w0lek:
This PR is a result of qt6 branch review.
The PR fix following issues:

issue1: stray issue
Screenshot from 2023-11-09 17-07-54

issue2: TIMING_MISSING handling is missing compare to the qt5 variant
Screenshot from 2023-11-09 11-20-47

issue3: setup size and position using QDesktopWIdget part is missing on qt6 variant
Screenshot from 2023-11-09 11-59-41

Screenshot from 2023-11-09 12-04-37

issue4:
found using sequential_map instead of QMap in several places explicitly for qt6 branch, so i reviewed sequential_map class as well.
Screenshot from 2023-11-09 12-12-08

sequential_map in push_back() method doesn't check if the key already is presented, this allow to store multiple values with same key. but methods operator[] and value() return oldest(first key occurred) value, might be not actual. Maybe it's not issue with current data set, but better avoid having multi keys behavior, as it's not obvious from the sequential_map interface.


src/Utils/sequential_map.h
```diff
@@ -49,7 +49,20 @@ class sequential_map {
     return defaultValue;
   }
 
-  void push_back(const std::pair<Key, Value> &p) { m_data.push_back(p); }
+  void push_back(const std::pair<Key, Value> &p)
```
KochynVolodymyr, line 52: This is not related to Qt6

src/DesignRuns/create_runs_dialog.cpp
```diff
@@ -13,6 +15,13 @@ CreateRunsDialog::CreateRunsDialog(QWidget *parent)
   setWindowFlags(Qt::Dialog | Qt::WindowCloseButtonHint);
   m_formIndex = CRFI_SELECTTYPE;
 
+  // size and positioning dialog
+  QScreen *screen = QGuiApplication::primaryScreen();
```
KochynVolodymyr, line 19: How this related to Qt6 migration?

src/DesignRuns/device_planner_dialog.cpp
```diff
@@ -13,6 +15,13 @@ DevicePlannerDialog::DevicePlannerDialog(QWidget *parent)
   setWindowFlags(Qt::Dialog | Qt::WindowCloseButtonHint);
   setWindowTitle(tr("Device Planner"));
 
+  // size and positioning dialog
+  QScreen *screen = QGuiApplication::primaryScreen();
```
KochynVolodymyr, line 19: This code was removed on purpose since it is obsolete

src/Utils/sequential_map.h
```diff
@@ -49,7 +49,20 @@ class sequential_map {
     return defaultValue;
   }
 
-  void push_back(const std::pair<Key, Value> &p) { m_data.push_back(p); }
+  void push_back(const std::pair<Key, Value> &p) 
+  { 
+    // remove previosly added pair with same key, to escape sitation of having multiple keys
+    for (auto it=m_data.begin(); it != m_data.end();) {
```
KochynVolodymyr, line 55: Please fix formatting

src/DesignRuns/create_runs_dialog.cpp
w0lek: got it, i removed it

src/DesignRuns/device_planner_dialog.cpp
w0lek: got it, i removed it

w0lek:
new PR as a replacement is created https://github.com/os-fpga/FOEDAG/pull/1394

PR #1402, opened 2023-12-02
Macos qt6 issue

KochynVolodymyr:

Fixed MACOS build issue.
This PR fixes few different build issues:
* `#include <QtWidgets>` causes build fail
image

* Override virtual slot TaskTableView::dataChanged. This causes link error. 
image

This issue also present on Linux but for some reason does not cause link error. Command for detect problem:
```
find dbuild/ -name mocs_compilation.cpp.o | xargs -n 1 -I % sh -c 'objdump -x % | grep registerConverterFunction'
```

Error output:
````
0000000000000000         *UND*	0000000000000000 _ZN9QMetaType27unregisterConverterFunctionES_S_
0000000000000000         *UND*	0000000000000000 _ZN9QMetaType25registerConverterFunctionERKSt8functionIFbPKvPvEES_S_
0000000000000023 R_X86_64_PLT32    _ZN9QMetaType27unregisterConverterFunctionES_S_-0x0000000000000004
0000000000000045 R_X86_64_PLT32    _ZN9QMetaType25registerConverterFunctionERKSt8functionIFbPKvPvEES_S_-0x0000000000000004
````

There are two function that are not resolved. `unregisterConverterFunction` and `registerConverterFunction`. But why this issue reproducible on MACOS only? also there are many other undefined functions but this doesn't cause any problem.  

This issues leads to link error for `test_install` target
image

1. Wrong usage of header guards
2. Undefined link issue related to type conversion for QVariant::fromValue
3. qRegisterMetaType<FOEDAG::TaskStatus>("FOEDAG::TaskStatus"); 

 - [x] Library: compiler, foedagcore
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,7 @@ set(VERSION_MINOR 0)
 # Add the spdlog directory to the include path
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/third_party/spdlog/include ${CMAKE_CURRENT_SOURCE_DIR}/third_party/exprtk) 
 
-set(VERSION_PATCH 295)
+set(VERSION_PATCH 296)
 
 
 option(
--- a/src/Command/Command.h
+++ b/src/Command/Command.h
@@ -18,14 +18,13 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+#ifndef COMMAND_H
+#define COMMAND_H
 
 #include <fstream>
 #include <iostream>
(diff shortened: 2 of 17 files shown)


ravic-rs:
@KochynVolodymyr thanks for fixing this.
Please not that in the describe technical details, we need to post the problem and reasoning behind it.
It won't be clear always from reading the code.
These are for future macOS/other build reasons.

KochynVolodymyr:
@ravic-rs there no much information. This issues were fixed in experimental way and I don't have a real reason behind. But I put all info I have.

PR #1406, opened 2023-12-08
Fix parent block

ManadherRS:
> - Fix Instance addition scope

### DIFF
--- a/src/DeviceModeling/device_modeler.h
+++ b/src/DeviceModeling/device_modeler.h
@@ -966,11 +966,11 @@ class device_modeler {
     if ("" != logic_location_z) {
       logic_location_z_i = convert_string_to_integer(logic_location_z);
     }
-    block->instance_vector().push_back(std::make_shared<device_block_instance>(
+    parent_block->instance_vector().push_back(std::make_shared<device_block_instance>(
         block, block->instance_vector().size(), logic_location_x_i,
         logic_location_y_i, logic_address_i, name, io_bank,
         logic_location_z_i));
-    block->instances()[name] = block->instance_vector().back();
+    parent_block->instances()[name] = block->instance_vector().back();
     return true;
   }
   /**
--- a/src/DeviceModeling/device_modeler.h
+++ b/src/DeviceModeling/device_modeler.h
@@ -966,10 +966,11 @@ class device_modeler {
     if ("" != logic_location_z) {
       logic_location_z_i = convert_string_to_integer(logic_location_z);
     }
-    parent_block->instance_vector().push_back(std::make_shared<device_block_instance>(
-        block, block->instance_vector().size(), logic_location_x_i,
-        logic_location_y_i, logic_address_i, name, io_bank,
-        logic_location_z_i));
+    parent_block->instance_vector().push_back(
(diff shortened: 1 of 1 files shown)


src/DeviceModeling/device_modeler.h
```diff
@@ -1077,11 +1077,12 @@ class device_modeler {
...
-        block, block->instance_vector().size(), logic_location_x_i,
-        logic_location_y_i, logic_address_i, name, io_bank,
-        logic_location_z_i));
-    block->instances()[name] = block->instance_vector().back();
+    parent_block->instance_vector().push_back(
+        std::make_shared<device_block_instance>(
+            block, block->instance_vector().size(), logic_location_x_i,
+            logic_location_y_i, logic_address_i, name, io_bank,
+            logic_location_z_i));
+    parent_block->instances()[name] = block->instance_vector().back();
```
chungshien, line 1085: You should not just directly overwrite the instance_map_ (or instances()) without giving any error to user.

There is function you should call add_instance() - https://github.com/os-fpga/FOEDAG/blob/main/src/DeviceModeling/device_block.h#L482-L487
- it did the checking
- it throw error if the infra is misused

chungshien: Beside the suggested add_instance(), there are more issue in these changes. It will cause invalid memory access.

I will create new PR with proper fix, plus other issues (more on parameter)

alain-rs:
@chungshien , lets close this PR. It has bugs.

ManadherRS:
@chungshien Can you explain , please?
 <there are more issue in these changes. It will cause invalid memory access.>

ManadherRS:
@alain-rs my changes passed all tests : https://github.com/os-fpga/FOEDAG/actions/runs/7135610160

chungshien:
image

1. You always use the wrong size as the instance ID
2. You will access something that might not even exist. If you truly run a real RIC TCL which you could find it in our tool, you should get "Segmentation fault" error

image

ManadherRS:
That's not true !
In the execution of C++ functions, the parameters are executed before the function itself, so the evaluation of the vector size happens before the build of the instance thus before the execution of the push_back and the update of the vector size.
In fact this is a widely used pattern in vector handling. Please , give it a try, it's a cool trick.
@alain-rs @chungshien

ManadherRS:
Let me do this, I'll split the JIRA case on the correction of the Parent model and the check for duplicate name .
Then I'll submit separate fixes ...
I'll put you as reviewer and we discuss the issue on code...

alain-rs:
@ManadherRS let @chungshien  finish this task.
You have other higher priorities

ManadherRS:
Done with this. (The code was ready, so I checked it in)
I'll answer any question ...

chungshien:
?!?!?!?!?

Is it this very obvious.

You add a new item to **parent_block**'s instance_vector @ line 969. However **block**'s instance_vector is still empty

@ line 974, you try to access last item of **block**'s instance_vector (which is empty, you should access **parent_block**), hence causing segmentation error!!!!!

ManadherRS:
Here's an example program to prove you wrong : http://tpcg.io/_F7KA3K

PR #1408, opened 2023-12-08
Fix the Parent Block pointer

ManadherRS:
> -  Fix the parent block pointer of an instance in RIC


alain-rs:
@chungshien , I'm merging this.
Please add the call to add_instance() in a subsequent PR.

PR #1417, opened 2023-12-20
[RG-306] Code improvement to avoid conversion from string to string list for project mananger

KochynVolodymyr:
 - [x] To address an existing issue. If so, please provide a link to the issue: RG-306

1. Use string list instead of conversion string to string list for design and simulation files. Before the change, files list was compiled into the string with space separator. Inside ProjectManager this string splits into list of strings. After changes, there is no string for files but only string list. 
2. Encapsulate conversion between `std::string` and `QString` inside `TclCommandIntegration` class. Provide interface with `std::string` only for compiler. 

 - [x] Library: compiler, newproject, projnavigator, utils
 - [ ] Engine
 - [ ] Documentation
 - [x] Regression tests


ravic-rs:
@KochynVolodymyr do we need to change anything upstream for this refactoring ?

KochynVolodymyr:
@ravic-rs No, we don't need

PR #1419, opened 2023-12-21
Reconstruct verilog exe integrtion

nadeemyaseen-rs:
Integrating reconstruct verilog executable.


nadeemyaseen-rs:
@ManadherRS As Alain suggested, I have renamed the executable. The executable execution process is the same as the VPR executable execution. Let me know if any other changes are required else we can proceed to merge.

PR #1435, opened 2024-01-03
Set LD LIBRARY PATH in CentOS for OpenSSL 

nadeemyaseen-rs:


.github/workflows/main.yml
```diff
@@ -213,14 +213,14 @@ jobs:
     - name: Test/batch
       if: matrix.mode == 'test/batch'
       run: |
-        export LD_LIBRARY_PATH=${GITHUB_WORKSPACE}/dbuild/bin/gtkwave/lib:$LD_LIBRARY_PATH
+        export LD_LIBRARY_PATH=${GITHUB_WORKSPACE}/dbuild/bin/gtkwave/lib:${GITHUB_WORKSPACE}/third_party/openssl:$LD_LIBRARY_PATH
```
KochynVolodymyr, line 216: Does it work with installed foedag?

nadeemyaseen-rs: On CentOS 7 CI, there is no install foedag.  Also when foedag will be installed on CentOS 7 then on CI we will install those libs too. So it will work.

KochynVolodymyr (approved):

PR #1449, opened 2024-01-11
Temporary disable test_install target for macos due to build issue

KochynVolodymyr:
temporary disable macos test_install target

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -462,9 +462,9 @@ jobs:
         make release
         make install
 
-    - name: Unit tests
-      run: |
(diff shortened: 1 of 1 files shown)


ravic-rs:
@KochynVolodymyr my understanding was that we will use a flag to disable this just for macos.

KochynVolodymyr:
@ravic-rs This is disabled just for macos. But flags can't be used in this case.

ravic-rs:
> @ravic-rs This is disabled just for macos. But flags can't be used in this case.

Ahh got it, thanks!

PR #1478, opened 2024-02-16
Update version.yml to solve permission problem for increment.yml

NadeemYaseen:
Updated version.yml to provide sufficient permission to push back to repo


NadeemYaseen:
It is now passing: https://github.com/os-fpga/FOEDAG/actions/runs/7931106564/job/21654646314

PR #1491, opened 2024-03-06
Rename nlohmann_json target to foedag_nlohmann_json to avoid name clash

NadeemYaseen:
Since nlohmann_json is used in other open-source projects, we have a name clash in the CMake target so just appending the prefix foedag to avoid the name clash.


behzadmehmood (approved):

NadeemYaseen:
CI is passed: https://github.com/os-fpga/FOEDAG/actions/runs/8169864173

PR #1505, opened 2024-03-13
Commented the section of setting verilator root

NadeemYaseen:
Now Verilator supports the relocation of binaries so no longer a need to set the VERILATOR_ROOT environment variable before invoking the verilator.


NadeemYaseen:
Along these changes, an update in the path of the verilator bin folder is also required so when bumping this change to FOEDAG_rs, use branch [verilatorpatch](https://github.com/os-fpga/FOEDAG_rs/tree/verilatorpatch). This branch already has this change.

PR #1514, opened 2024-03-22
Pass base device name to license check in case of custum device

NadeemYaseen:
> - [ ] Breaking new feature. If so, please describe the details in the description part.

When the end user creates a custom device, pass the base device name to verify the license availability.

### DIFF
--- a/src/Compiler/CompilerOpenFPGA.cpp
+++ b/src/Compiler/CompilerOpenFPGA.cpp
@@ -3630,13 +3630,6 @@ bool CompilerOpenFPGA::LoadDeviceData(
   if (!deviceFound) {
     status = false;
   }
-  /*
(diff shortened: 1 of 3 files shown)


alaindargelas:
@ravikiranchollangi , please push this through to raptor, build, test with and without custom device in production build.

src/Compiler/CompilerOpenFPGA.cpp
```diff
@@ -3586,6 +3592,7 @@ bool CompilerOpenFPGA::LoadDeviceData(
                   status = false;
                 }
               } else if (file_type == "base_device") {
+                BaseDeviceName(name);
                 // field is used for identify base for custom device
                 // no action so far
```
volodymyrkochyn, line 3597: this comment need to remove

PR #1535, opened 2024-04-09
Account for io row/column

alaindargelas:

The vpr layout has additional rox/columns for IOs.
This abstract that away.
Row,Column numbers are real fabric numbers now. Resource count is corrected.
Offsets for DSP/BRAM are also adjusted.


volodymyrkochyn:
@alaindargelas here is a hint which explain how the values calculated. Hint does not updated

PR #1563, opened 2024-05-02
Fix the include error for scope_guard.hpp

NadeemYaseen:
The scope_guard.hpp directory was not in the include directory so added it in the include directory so that compiler can find it.


alaindargelas (approved):

ravikiranchollangi:
Thanks @NadeemYaseen , the CI got stuck with status in this one. Hoping it's not a action version mismatch now.

PR #1584, opened 2024-05-15
fix .pin to .pcf translation for pin_c EDA-2857

serge-dsa:


alaindargelas:
As discussed, we need a checker, this use case is not allowed in the HW.
You can submit another PR. Some of the diagnostic code in this PR can be ported to the new PR.

PR #1585, opened 2024-05-15
reject set_prop command if it overwrites existing mode EDA-2857

serge-dsa:


alaindargelas:
@serge-dsa , please fix the code formatting:
.github/bin/run-clang-format.sh

PR #1606, opened 2024-05-31
Load data from config.json after project file loaded. This will fix clock names in the reports

volodymyrkochyn:

Load data from config.json after project file loaded. This will fix clock names in the reports
image

 - [x] Library: foedagcore
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests

### DIFF
--- a/src/Main/ProjectFile/CompilerComponent.cpp
+++ b/src/Main/ProjectFile/CompilerComponent.cpp
@@ -51,4 +51,10 @@ ErrorCode CompilerComponent::Load(QXmlStreamReader *reader) {
   return {};
 }
 
+void CompilerComponent::LoadDone() {
+  std::filesystem::path configJsonPath =
+      m_compiler->FilePath(Compiler::Action::Synthesis) / "config.json";
+  m_compiler->getNetlistEditData()->ReadData(configJsonPath);
+}
+
 }  // namespace FOEDAG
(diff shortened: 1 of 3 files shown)


src/Main/ProjectFile/CompilerComponent.cpp
```diff
@@ -51,4 +51,10 @@ ErrorCode CompilerComponent::Load(QXmlStreamReader *reader) {
   return {};
 }
 
+void CompilerComponent::LoadDone() {
+  std::filesystem::path configJsonPath =
```
alaindargelas, line 55: If the file does not exist (Synthesis didn't execute yet), this should not be run. Test if file exist.

alaindargelas (changes requested):

src/Main/ProjectFile/CompilerComponent.h
```diff
@@ -32,6 +32,7 @@ class CompilerComponent : public ProjectFileComponent {
   CompilerComponent(Compiler *cc);
   void Save(QXmlStreamWriter *writer) override;
   ErrorCode Load(QXmlStreamReader *reader) override;
+  void LoadDone() override;
```
alaindargelas, line 35: Where is this called?

alaindargelas (changes requested):

src/Main/ProjectFile/CompilerComponent.h
volodymyrkochyn: This called when project file loaded

PR #1621, opened 2024-06-14
Add Run TCL script action to allow run script from the UI

volodymyrkochyn:

Run TCL script, new feature that allow user to execute tcl script from the GUI. User action can be found under File->Run TCL script..
The flow is:
1. Ask user if close previous project
2. Allow user select tcl script
3. Run selected script 

image

 - [x] Library: foedagcore
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests


ravikiranchollangi (changes requested):
Do you think we should move this to the "Run" menu ? It seems appropriate to be there at first. But I also think because it runs a new script/creates new project, it might not belong there.

volodymyrkochyn:
> Do you think we should move this to the "Run" menu ? It seems appropriate to be there at first. But I also think because it runs a new script/creates new project, it might not belong there.

This is something in between. It creates new project (in most cases but it could be something else) and close the old one, open some file. For me it more close to the File menu. But at the same time it starts compilation.  I don't have strict opinion on that.

ravikiranchollangi (approved):

PR #1645, opened 2024-07-04
Pass device name and device file path to IP catalog

volodymyrkochyn:

Change list:
* Extend IP Catalog communication protocol with two new fields:
  * device - device name
  * device_path - path to device file.
* Show warning message when user switch target device

image
 

 - [x] Library: newproject, compiler, ipconfigurator
 - [ ] Engine
 - [ ] Documentation
 - [ ] Regression tests


ravikiranchollangi:
We should also put up the status on top right which tells user that the design has changed and needs recompile. Right ?

volodymyrkochyn:
> We should also put up the status on top right which tells user that the design has changed and needs recompile. Right ?

Right. We have that status.

PR #1670, opened 2024-08-09
Add top module for VHDL with multiple entities

awaisabbas006:
> - [ ] If there are multiple VHDL entities, then GHDL errors out with no top module found
> - [ ] Top module for VHDL design is added with option "-e top_module"


awaisabbas006:
@alain-rs please review the PR.

PR #1679, opened 2024-08-21
Partially revert recent CGA test changes due to local testing issues

abdulhameed-akram:
Reverting specific changes from the recent commit that caused issues in local testing after merging the FOEDAG tests for CGA. Some modifications will remain as they do not affect.


ravikiranchollangi:
@abdulhameed-akram please verify these on CGA locally before requesting changes here. I assumed in previous PR that these were tested already.

PR #1708, opened 2024-09-26
add use_vpr_latest command EDA-3111

serge-dsa:


src/Compiler/CompilerOpenFPGA.cpp
```diff
@@ -492,6 +492,25 @@ bool CompilerOpenFPGA::RegisterCommands(TclInterpreter* interp,
...
+    std::filesystem::path vpr = comp.m_vprExecutablePath.lexically_normal();
+    constexpr bool dbg = false;
+    if (dbg) std::cerr << "\n  VPR WAS: " << vpr << '\n';
+    try {
+      vpr.replace_filename("vpr_latest");
+      comp.VprExecPath(vpr);
+      if (dbg) std::cerr << "  VPR NOW: " << vpr << "\n\n";
+    } catch (...) {
+      if (dbg) {
+        assert(0);
```
ravikiranchollangi, line 507: Why do we need a assert and a dgb flag ?

src/Compiler/CompilerOpenFPGA.cpp
```diff
@@ -492,6 +492,25 @@ bool CompilerOpenFPGA::RegisterCommands(TclInterpreter* interp,
   };
   interp->registerCmd("flat_routing", flat_routing, this, 0);
 
+  auto use_vpr_latest = [](void* clientData, Tcl_Interp* interp, int argc,
+                           const char* argv[]) -> int {
+    CompilerOpenFPGA& comp = *(CompilerOpenFPGA*)clientData;
+    std::filesystem::path vpr = comp.m_vprExecutablePath.lexically_normal();
+    constexpr bool dbg = false;
+    if (dbg) std::cerr << "\n  VPR WAS: " << vpr << '\n';
```
ravikiranchollangi, line 500: Please use this instead.
https://github.com/os-fpga/FOEDAG/blob/main/src/Compiler/CompilerOpenFPGA.cpp#L2906C7-L2906C19

ravikiranchollangi (commented):
2 comments

src/Compiler/CompilerOpenFPGA.cpp
serge-dsa: replace_filename may throw exceptions according to
https://en.cppreference.com/w/cpp/filesystem/path/replace_filename

assert(0) make it easy to debug the exception situation

'dbg' is not really a flag, it's a way to comment out code. It's constexpr false at compile time so the parts after if(dbg) are not compiled, it's equivalent to commenting out. I'll remove this stuff.

src/Compiler/CompilerOpenFPGA.cpp
```diff
@@ -492,6 +492,20 @@ bool CompilerOpenFPGA::RegisterCommands(TclInterpreter* interp,
...
 
+  auto use_vpr_latest = [](void* clientData, Tcl_Interp* interp, int argc,
+                           const char* argv[]) -> int {
+    CompilerOpenFPGA& comp = *(CompilerOpenFPGA*)clientData;
+    std::filesystem::path vpr = comp.m_vprExecutablePath.lexically_normal();
+    try {
+      vpr.replace_filename("vpr_latest");
+      comp.VprExecPath(vpr);
+    } catch (...) {
+      assert(0);
```
alaindargelas, line 503: Asserts are easy for developers to debug.
They are a nightmare for end-users.
Please use the ErrorMessage instead and print a proper message like could not find this executable and return Tcl Error, so the Tcl script stops accordingly with the proper error trace.

alaindargelas (changes requested):

src/Compiler/CompilerOpenFPGA.cpp
serge-dsa: @alaindargelas assert() is enabled only in 'make debug' so the end users won't experience it.
replace_filename is a purely lexical (string) operation, it does not care if the paths exist.
I don't know what kind of exceptions it can throw, but cppreference and other references say it can, potentially.

replace_extension() is already used in other places in FOEDAG without catching exceptions, like some other C++ utilities that may throw are used without checking exceptions in FOEDAG. This assert(0) and exception is extremely unlikely case, I'll remove the try-catch to be consistent with FOEDAG code :).

src/Compiler/CompilerOpenFPGA.cpp
ravikiranchollangi: Got it, thanks.
We don't have a way to handle debug messages yet. You can use the ErrorMessage option as requested.

alaindargelas (approved):

